Issue #6, opened 2019-02-21
A full non-relative URL in module-shims?


Hypercubed:
Is there a technical reason that full non-relative URLs don't work inside shimmed modules?

For example:

```
<script defer src="https://unpkg.com/es-module-shims@0.2.0/dist/es-module-shims.js"></script>
<script type="importmap-shim">
{
  "imports": {
    "lit-html": "https://unpkg.com/lit-html@1.0.0/lit-html.js"
  }
}
</script>
<script type="module-shim">
  import { html } from "lit-html";                                             // works
  // import { html } from "https://unpkg.com/lit-html@1.0.0/lit-html.js";      // does not work
  console.log(html);
</script>
```

Codepen for testing: https://codepen.io/Hypercubed/pen/daLrMZ

The error is:
```
Uncaught (in promise) Error: Unable to resolve bare specifier "https://unpkg.com/lit-html@1.0.0/lit-html.js" from https://s.codepen.io/boomerang/iFrameKey-d5cb004a-1e37-cbae-c83b-d4327172eec9/
    at throwBare (es-module-shims.js:174)
    at resolveImportMap (es-module-shims.js:170)
    at resolve (es-module-shims.js:925)
    at load.d.Promise.all.deps.map (es-module-shims.js:880)
    at Array.map (<anonymous>)
    at load.L.load.f.then (es-module-shims.js:879)
```

Hypercubed:
This also throws an error:

```
importShim('https://unpkg.com/lit-html@1.0.0/lit-html.js').then(x => console.log(x));
```

bkardell:
Using the code from /dist right now this seems to work for me?

https://bramble-veil.glitch.me/other.html

Hypercubed:
@bkardell I think your demo is not the same... importing a full non-relative URL (with https://) works via the import map... but not when the URL is used directly.

bkardell:
Ah, yes, I see... This one captures it? https://bramble-veil.glitch.me/top.html

As I understand it, _most_ things may not be in the import map and the idea is that they would then 'fall through', but it looks like there is no falling through unaccosted in `resolveIfNotPlainOrUrl(..)`.

[...]

PR #7, opened 2019-02-21
support non-relative URLs with protocol

Hypercubed:
Potential fix for #6 

* Supports `http(s):`, `data:`, and `blob:` urls.
* Adds tests for `http(s):`, `data:`, and `blob:` urls.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -12,17 +12,12 @@ if (typeof location !== 'undefined') {
 }
 
 const backslashRegEx = /\\/g;
-const protocolre = /^[a-z][a-z0-9.+-]*\:/i
 export function resolveIfNotPlainOrUrl (relUrl, parentUrl) {
   if (relUrl.indexOf('\\') !== -1)
     relUrl = relUrl.replace(backslashRegEx, '/');
   // protocol-relative
   if (relUrl[0] === '/' && relUrl[1] === '/') {
     return parentUrl.slice(0, parentUrl.indexOf(':') + 1) + relUrl;
-  }
-  // non-relative URL with protocol
-  else if (protocolre.test(relUrl)) {
-    return relUrl;
   }
   // relative-url
   else if (relUrl[0] === '.' && (relUrl[1] === '/' || relUrl[1] === '.' && (relUrl[2] === '/' || relUrl.length === 2 && (relUrl += '/')) ||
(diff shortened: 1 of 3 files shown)


Hypercubed:
The test for non-relative URL with the protocol is a check for '://':

* this may need to be modified to support specific protocols
* may need to support `blob:` and `data:`.

bkardell:
Deleted my comment as i had this window open for hours and I see now that you'd made a pull already. Apologies for the noise.

src/common.js
```diff
@@ -12,13 +12,18 @@ if (typeof location !== 'undefined') {
...
   if (relUrl.indexOf('\\') !== -1)
     relUrl = relUrl.replace(backslashRegEx, '/');
   // protocol-relative
   if (relUrl[0] === '/' && relUrl[1] === '/') {
     return parentUrl.slice(0, parentUrl.indexOf(':') + 1) + relUrl;
   }
+  // non-relative URL with protocol
+  else if (protocolre.test(relUrl)) {
+    return relUrl;
+  }
```
guybedford: Thanks so much for posting this. The fix is right, but I would rather move this check to just before its usage in the import maps implementation rather. This resolve function was carefully designed around the WhatWG specification of relative URL matching - I'm actually against supporting maps for URLs so am still secretly hoping this gets cut from the spec for a better approach of handling the polyfill codepaths it was created for (ie this project!).

Hypercubed: Hi @guybedford Do you think this check belongs in the `resolve` function:

```js
const protocolre = /^[a-z][a-z0-9.+-]*\:/i
async function resolve (id, parentUrl) {
  if (protocolre.test(id)) {
    return Promise.resolve(id);
  }
  parentUrl = parentUrl || pageBaseUrl;

  if (importMapPromise)
    return importMapPromise
    .then(function () {
      return resolveImportMap(id, parentUrl, importMap);
    });

  return resolveImportMap(id, parentUrl, importMap);
}
```

Does it look like this is the equivalent for system.js: https://github.com/systemjs/systemjs/blob/60cd9e2c3d2d7c3765c9afa10b5bd205ed7a71d9/src/features/basic-resolve.js#L5 ?

Hypercubed: Although it looks like the resolve is the same as when using importMaps: https://github.com/systemjs/systemjs/blob/60cd9e2c3d2d7c3765c9afa10b5bd205ed7a71d9/src/features/import-map.js#L39

guybedford: I would include it as a check in the resolveImportMap function if the return value from `resolveIfNotPlainOrURL` is not defined - https://github.com/systemjs/systemjs/blob/60cd9e2c3d2d7c3765c9afa10b5bd205ed7a71d9/src/common.js#L162.

bkardell: so like https://glitch.com/edit/#!/bramble-veil?path=shimport/es-module-shims-temp.js:165:16 ?

guybedford: Yes exactly!

Hypercubed: Fixed!

guybedford:
Looks great, thanks for the PR!

PR #17, opened 2019-06-13
Add support for import maps in Worker; resolves #9

### LINKED ISSUE #9: Expose import map shim on JS API
Currently, shim only supports defining import maps via dom.

`<script type="importmap-shim">`

We can `src` or `inline` the map, but there is a hard constraint on the existence of this element in the dom.

Would be nice to be able to use the shim in a web worker (which doesn't have a dom).

Basically, import the shim into a webworker and call something like `useMap(url|object)`

I understand the shim is trying to force a pattern as close to the final spec as possible... maybe the lib could detect it is in a webworker before defining/exposing the function?  Or throw an exception if used on the main thread. 

Happy to discuss further...

costingeana:
Add support for import maps in Worker. Resolves #9. 
1. add WorkerShim class;
2. 'expose' importMap on global object (self) in web workers and browser.


guybedford:
Amazing work and thank you for the PR! I really like how simple this turned out to be.

A nice addition might be to add support for `new Worker` detection in the lexer and handle the rewriting automatically without needing to use `WorkerShim`, but I'd be happy to include that as a possible future follow-on to this PR.

@costingeana would you be able to include a simple test before we land this?

costingeana:
Thank you @guybedford, but I only followed your initial idea from https://github.com/guybedford/es-module-shims/issues/9#issuecomment-476910287. 
I just needed import maps in Workers in a project I'm working on and I used your `WorkerShim` proof of concept, which worked beautifully.

What's next?
Firstly I would add some tests for the current implementation.
Secondly I will implement and test your suggestion: `new Worker` detection in lexer.

Thank you for the feedback on this pull request; it helps a lot.

guybedford:
@costingeana happy to merge here with just a test, then let's work on the lexer addition in a new PR. The lexer doesn't have any concept of identifiers and referenced identifiers vs other types of identifiers, but we can probably just work with the fact that the `new` keyword followed by the `Worker` identifier would only ever be valid in the right place (eg `function (new Worker)` isn't supported etc). So again we can hack around the limitations of the lexer in a well defined way hopefully here!

costingeana:
My apologies for the delay!
@guybedford, please let me know if there is something else I can add (or remove).

costingeana:
@guybedford, is there anything else I can do to make this pull request ready for merging?

test/test.html
```diff
@@ -19,7 +19,7 @@
   import test from "test";
   console.log(test);
 </script>
-<script type="module" src="../src/es-module-shims.js"></script>
+<script type="module" src="../dist/es-module-shims.js"></script>
```
guybedford: Was this necessary for the tests to pass? I must admit I quite like being able to run the tests without a rebuild.

costingeana: Short answer: yes, this is necessary for tests to pass.
Long answer: The `importScript()` method used in WorkerShim constructor fails if the script it loads has ES6 syntax; as far as I know the `importScripts()` method will automatically fail inside module workers.
```
const workerScriptUrl = createBlob(\`importScripts('${es_module_shims_src}'); 
    self.importMap = ${JSON.stringify(options.importMap || {})}; importShim('${new URL(aURL, pageBaseUrl).href}')`);
```

dist/es-module-shims.js
```diff
@@ -689,6 +693,32 @@
...
+      for (let i = 0, len = scripts.length; i < len; i++) {
+        if (scripts[i].src.includes('es-module-shims.js')) {
+            es_module_shims_src = scripts[i].src;
+
+            break;
+        }
+      }
+
+      const workerScriptUrl = createBlob(`importScripts('${es_module_shims_src}'); 
+    self.importMap = ${JSON.stringify(options.importMap || {})}; importShim('${new URL(aURL, baseUrl).href}')`);
```
guybedford: Indentation.

test/test.html
guybedford: Ah I see what you mean, thanks.

Let's go with (2) then so that this test restriction only applies to the worker tests.

src/worker-shims.js
```diff
@@ -0,0 +1,27 @@
...
+
+    let es_module_shims_src = new URL('es-module-shims.js', pageBaseUrl).href;
+    const scripts = document.scripts;
+    for (let i = 0, len = scripts.length; i < len; i++) {
+      if (scripts[i].src.includes('es-module-shims.js')) {
+          es_module_shims_src = scripts[i].src;
+
+          break;
+      }
+    }
```
guybedford: Rather than this restriction, could we instead try something like - 

```js
let esModuleShimsSrc;
if (document.currentScript)
  esModuleShimsSrc = document.currentScript.href;
```

Then in the `WorkerShim` constructor, something like - 

```js
if (!esModuleShimsSrc) throw new Error('es-module-shims.js must be loaded with a script tag for WorkerShim support.');
```

note the above would have to be outside of the class constructor and instead in the script initialization.

dist/es-module-shims.js
costingeana: I think I submitted by mistake this file, namely `dist/es-module-shims.js`. I shouldn't have done that. I apologize.

guybedford:
Amazing work, thank you!

costingeana:
You’re welcome! And thank you, too.

By the end of this week I will come up with another pull request to add support for new Worker detection in the lexer as you suggested earlier.

Issue #24, opened 2019-06-26
Object spread


LarsDenBakker:
Object spread is used here: https://github.com/guybedford/es-module-shims/blob/master/src/worker-shims.js#L15 which is not supported on all browsers yet. Specifically Edge doesn't support it.

Replacing it with `Object.assign` seems like a good trafeoff.

PR #25, opened 2019-06-26
Use Object.assign instead of object spread

LarsDenBakker:
Fixes https://github.com/guybedford/es-module-shims/issues/24


guybedford:
Thank you for the PR - published 0.2.11.

PR #26, opened 2019-06-28
Take base href into account when resolving relative urls

LarsDenBakker:
The same change as: https://github.com/systemjs/systemjs/pull/1957 to solve the issue described here: https://github.com/systemjs/systemjs/issues/1956

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -4,7 +4,14 @@ const envGlobal = hasSelf ? self : global;
 export { envGlobal as global };
 
 export let baseUrl;
-if (typeof location !== 'undefined') {
+
+if (typeof document !== 'undefined') {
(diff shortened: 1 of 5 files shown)


dist/es-module-shims.js
```diff
@@ -3,7 +3,14 @@
   'use strict';
 
   let baseUrl;
-  if (typeof location !== 'undefined') {
+
+  if (typeof document !== 'undefined') {
+    const baseEl = document.head.querySelector('base[href]');
```
guybedford: To be incredibly pedantic, I think we can save a byte or two by making this `document.querySelector('base[href]')`.

LarsDenBakker:
Made the changes :)

guybedford:
Released in 0.2.12.

Issue #19, opened 2019-06-22
Fallback support or note in documentation


begrs:
Since fallbacks in import maps are not supported yet, please either implement them or add a note to the Readme to make it obvious to devs that this part of import maps is not supported.

I wanted to use import shims mainly for that purpose so I am a little disappointed that this basic/important feature when working with cdn provided js files is not yet implemented.

best reagards

guybedford:
As specified currently, import maps fallbacks only apply to builtin modules using the `std:` native namespace in the browser itself.

Fetch scheme fallbacks have not been specified, and personally I hope and will strongly argue for them not to be a feature.

So the best we could do to support this right now while remaining spec compliant would be to always treat `"imports": { "x": ["std:x", "/x.js"] }` as resolving `/x.js`.

I'd be glad to take a PR along those lines.

guybedford:
This has been implemented.

PR #28, opened 2019-07-06
support fallback imports from import map

thepassle:
Regarding this issue: https://github.com/guybedford/es-module-shims/issues/19 Would this suffice to support fallbacks for now? This should always resolve the fallback instead of the built-in module, as you suggested in that issue. Please let me know your thoughts on this :-) Would happily take some pointers

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -123,9 +123,11 @@ function resolvePackages(pkgs) {
   for (var p in pkgs) {
     var value = pkgs[p];
     // TODO package fallback support
-    if (typeof value !== 'string')
-      continue;
-    outPkgs[resolveIfNotPlainOrUrl(p) || p] = value;
+    if (typeof value !== 'string') {
+      outPkgs[resolveIfNotPlainOrUrl(p) || p] = value.filter(v => !v.startsWith('std:'))[0];
+    } else {
+      outPkgs[resolveIfNotPlainOrUrl(p) || p] = value;
+    }
   }
   return outPkgs;
 }
--- a/src/common.js
+++ b/src/common.js
@@ -123,11 +123,10 @@ function resolvePackages(pkgs) {
   for (var p in pkgs) {
     var value = pkgs[p];
     // TODO package fallback support
-    if (typeof value !== 'string') {
-      outPkgs[resolveIfNotPlainOrUrl(p) || p] = value.filter(v => !v.startsWith('std:'))[0];
-    } else {
+    if (Array.isArray(value))
+      value = value.find(v => !v.startsWith('std:'));
+    if (typeof value === 'string')
       outPkgs[resolveIfNotPlainOrUrl(p) || p] = value;
-    }
   }
   return outPkgs;
 }
--- a/src/common.js
+++ b/src/common.js
@@ -172,7 +172,7 @@ function applyPackages (id, packages, baseUrl) {
 
 const protocolre = /^[a-z][a-z0-9.+-]*\:/i;
 export function resolveImportMap (id, parentUrl, importMap) {
-  const urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);
+  const urlResolved = resolveIfNotPlainOrUrl(id, parentUrl) || id.indexOf(':') !== -1 && id;
   if (urlResolved){
     id = urlResolved;
   } else if (protocolre.test(id)) { // non-relative URL with protocol
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -17,7 +17,6 @@ catch (e) {
       const topLevelBlobUrl = createBlob(
         `import*as m from'${blobUrl}';self.importShim.l=m;self.importShim.e=null`
       );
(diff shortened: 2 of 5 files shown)


src/common.js
```diff
@@ -123,9 +123,11 @@ function resolvePackages(pkgs) {
   for (var p in pkgs) {
     var value = pkgs[p];
     // TODO package fallback support
-    if (typeof value !== 'string')
-      continue;
-    outPkgs[resolveIfNotPlainOrUrl(p) || p] = value;
+    if (typeof value !== 'string') {
```
guybedford: This should rather be `if (Array.isArray(value))` with `else if (typeof value === 'string')`.

src/common.js
```diff
@@ -123,9 +123,11 @@ function resolvePackages(pkgs) {
...
+      outPkgs[resolveIfNotPlainOrUrl(p) || p] = value.filter(v => !v.startsWith('std:'))[0];
```
guybedford: `value.find(v => !v.startsWith('std:'))` seems shorter here?

Also in the case where the value is undefined we shouldn't be assigning. So perhaps rather something like:

```js
if (Array.isArray(value))
  value = value.find(v => !v.startsWith('std:'));
if (typeof value === 'string')
  outPkgs[resolveIfNotPlainOrUrl(p) || p] = value;
```

as the whole thing.

thepassle:
Actually after playing around with this a bit more, I don't think this change alone works. If I change `await importShim('std:built-in');` with import map:
```json
{
  "imports": {
    "std:built-in": ["std:built-in", "./built-in-fallback.js"]
  }
}
```

I still get the error:
```
Fetch API cannot load std:built-in. URL scheme must be "http" or "https" for CORS request.
```

guybedford:
The other thing to consider here is if it should attempt to use `std:` modules when they do exist in the browser too... we could have code that does that detection by injecting a module tag with the import and checking if it errors. So yes, more to think about indeed.

thepassle:
We can probably achieve that in a similar way as:
```js
      const s = document.createElement('script');
      s.type = 'module';
      s.src = topLevelBlobUrl;
      document.head.appendChild(s);
      return new Promise((resolve, reject) => {
        s.addEventListener('load', () => {
          document.head.removeChild(s);
          if (importShim.e)
            return reject(importShim.e);
          resolve(importShim.l);
        });
      });
```
right?

Also, do you have any pointers regarding to [this](https://github.com/guybedford/es-module-shims/pull/28#issuecomment-508939075)? It seems to be storing the correct module resolution (the fallback), but I'm not sure how to access that module resolution when called with `importShim('std:built-in')`?

guybedford:
I think it's a bug in the import map resolution to not map URLs. See the check in https://github.com/guybedford/es-module-shims/blob/master/src/common.js#L177.

I actually changed this in SystemJS here - https://github.com/systemjs/systemjs/blob/master/src/common.js#L167, so simply updating from that implementation might achieve it since these code paths are supposed to be in sync.

thepassle:
I currently implemented it as you mentioned in https://github.com/guybedford/es-module-shims/issues/19#issuecomment-504705742 , and added a test for it as well. Would be happy to hear your feedback or any notes you may have :-)

In another iteration we could change the logic so it'll try to get the native built-in module, and on error get the fallback instead.

LarsDenBakker:
built-in modules also implies import map support, so would there be any need to cover that in the shim at the moment? Perhaps in the future when we need to shim additional import map features not implemented in some browsers.

guybedford:
Note that Chrome doesn't yet support `scopes` in the import map, so the shim is still useful in those environments, yet there may be cases where true builtin modules would be wanted.

I think we need to approach each feature in a fine-grained way individually.

thepassle:
Should we move that discussion to a separate issue? 🙈

Also, PTAL @guybedford 😊

src/es-module-shims.js
```diff
@@ -167,6 +165,11 @@ function getOrCreateLoad (url, source) {
     s: undefined,
   };
 
+  if(url.startsWith('std:')) {
+    load.u = importShim.map.imports[url]
+    url = importShim.map.imports[url]
+  }
```
guybedford: This workaround shouldn't be needed - see https://github.com/guybedford/es-module-shims/pull/28#issuecomment-508943173 for how we should handle this.

Basically updating the resolveImportMaps function to support URL mappings, as in the SystemJS implementation, which was done in this commit - https://github.com/systemjs/systemjs/commit/a0d53c956b751c.

guybedford:
(and yes, happy to move feature detections / native module support in a separate issue!)

src/es-module-shims.js
thepassle: Fixed 😊

guybedford:
Great work, thanks for pushing this one through!

thepassle:
Thanks for the feedback!

guybedford:
Published in 0.2.14.

PR #30, opened 2019-07-17
fix importing modules with path segments

LarsDenBakker:
When you import a module with query parameters or hashes it incorrectly imports dependencies as it's trying to join the paths including the query parameters and hashes. Stripping them before resolving, just like with the page base path, fixes this.


guybedford:
Published in 0.2.14.

Issue #35, opened 2019-08-07
Expose lexer


LarsDenBakker:
Until import maps are more mainstream, I'd like to use the lexer serverside in https://github.com/open-wc/open-wc/tree/master/packages/es-dev-server to resolve imports before serving them to the browsers.

Is it possible to expose the lexer module so that it can be imported separately?

guybedford:
Maybe it's time to actually make the lexer its own package actually. It is useful outside of this project.

Would you like to set that up here? I don't mind where it lives.

LarsDenBakker:
Yeah I can do that, but where should the package live?

guybedford:
We could create a new shared org for it, or put it under an existing one
like ES Module Loader is SystemJS.

LarsDenBakker:
Seems like https://github.com/ModuleLoader is a good place for this

daKmoR:
just an idea...
new group `es-module-tools` and have a monorepo in there as `es-module-tools`...
at open-wc we have quite some stuff now which is actually not really web components related but more es module related... so we could consider moving some of them 👍 

PS: `es-tools` is sadly already taken

guybedford:
I went ahead and separated the lexer out into its own project at https://github.com/guybedford/es-module-lexer.

Work to ensure this project properly depends on that is still pending. A PR is welcome, otherwise I'll get to it when I'm next working on this project.

LarsDenBakker:
Thanks, I will close this one. If I have time on the weekend I can make a PR.

PR #36, opened 2019-08-11
Use the separated es-module-lexer

LarsDenBakker:
As discussed in https://github.com/guybedford/es-module-shims/issues/35, this uses the separated lexer module.

I removed the lexer tests as well.


package.json
```diff
@@ -7,8 +7,6 @@
     "build": "rollup ./src/es-module-shims.js -f iife -o dist/es-module-shims.js --banner \"/* ES Module Shims 0.2.13 */\"",
     "min": "cd dist && terser es-module-shims.js -c passes=2 -m --source-map --comments \"/ES Module Shims/\" -o es-module-shims.min.js",
     "footprint": "npm run build && npm run min && cat dist/es-module-shims.min.js | gzip -9f | wc -c",
-    "test": "mocha -r esm -u tdd test/unit.js",
-    "test:samples": "mocha -r esm -u tdd test/samples.js",
     "bench": "node -r esm bench",
```
guybedford: These benchmarks can also be removed as this is lexer-specific.

src/es-module-shims.js
```diff
@@ -1,5 +1,5 @@
 import { baseUrl as pageBaseUrl, parseImportMap, resolveImportMap, createBlob } from './common.js';
-import { analyzeModuleSyntax } from './lexer.js';
+import analyzeModuleSyntax from '../node_modules/es-module-lexer/lexer.js';
```
guybedford: Unfortunately it's not quite this simple as I did actually change the API for dynamic import in the process of splitting out the lexer into its own package.

For `import('asdf')`, instead of the values being `import.s` reflecting the start of the `import(` statement and `import.d` being the start of the `'asdf` part, I switched it around so that `source.slice(import.s, import.e)` is the `'asdf` specifier and `import.d` is the `import(` statement part.

test/test-base-href.html
```diff
@@ -2,8 +2,8 @@
 <head>
   <base href="/test/fixtures/es-modules/">
 </head>
-<link rel="stylesheet" type="text/css" href="../../..///node_modules/mocha/mocha.css"/>
-<script src="../../..///node_modules/mocha/mocha.js"></script>
+<link rel="stylesheet" type="text/css" href="../../../node_modules/mocha/mocha.css"/>
+<script src="../../../node_modules/mocha/mocha.js"></script>
```
guybedford: Thanks

src/es-module-shims.js
LarsDenBakker: Updated to the new syntax. Added a test for dynamic import as well :) Please double check

src/es-module-shims.js
```diff
@@ -128,7 +128,7 @@ async function resolveDeps (load, seen) {
       }
       // dynamic import
       else {
-        resolvedSource += source.slice(lastIndex, start) + 'importShim' + source.slice(start + 6, end) + JSON.stringify(load.r) + ', ';
+        resolvedSource += source.slice(lastIndex, dynamicImportIndex) + 'importShim(' + JSON.stringify(load.r) + ', ' + source.slice(start, end);
```
guybedford: So this was originally done this way because the first version of the lexer didn't have the ability to match the opening and closing brackets around the `import( )` expression, so instead if `importShim(idExpr, parentURL)`, it was converted into `importShim(parentURL, idExpr` so that it could work out.

Now that we do support this, it could be worth switching the `importShim` function back around in the definition above (`async function importShim (id` line), and then simplifying this line to:

```suggestion
        resolvedSource += source.slice(lastIndex, dynamicImportIndex + 6) + 'Shim(' + source.slice(start, end) + ', ' + JSON.stringify(load.r);
```

LarsDenBakker: You meant swapping the `arguments` check inside `async function importShim`, right?

guybedford: Yes exactly.

guybedford:
I included some extra code removal as well in the merge. Thanks for working on this.

PR #37, opened 2019-08-20
Fixes source mapping in Firefox.

### LINKED ISSUE #4: Source Map support in Firefox?
I have a module that lives at `<root>/apple.js`.  In it there is a `//# sourceMappingURL=apple.js.map`.  However, when the browser tries to load that sourcemap it can't find it.  I suspect it has to do with the whole `blob:...` stuff, but I don't understand it well enough to speculate beyond that.

```
Source map error: TypeError: apple.js.map is not a valid URL.
Resource URL: blob:http://localhost:8081/033273b5-1e9d-4506-8bb5-cd301e633116
Source Map URL: apple.js.map
```

note: Source maps appear to work in Chrome.

MicahZoltu:
I made this change locally per recommendation in #4 and it seems to resolve the problem.  I also tested in Chrome, which appeared to work for me.

Breakpoints in Firefox still don't work from within source maps, but I think that is a Firefox bug, not an issue with this library.

I recommend having either CI or another human manually test this change.  I have no idea what I'm doing and just took the code referenced in #4 and pasted it verbatim.

Fixes #4

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -132,6 +132,7 @@ async function resolveDeps (load, seen) {
     resolvedSource += source.slice(lastIndex);
   }
 
+  resolvedSource = resolvedSource.replace(/\/\/# sourceMappingURL=(.*)\s*$/, (match, url) => match.replace(url, new URL(url, load.r)));
   load.b = createBlob(resolvedSource + '\n//# sourceURL=' + load.r);
   load.S = undefined;
 }


guybedford:
I'm very glad to hear that this is working, and thanks for reporting back.

I will merge this in soon, I'd just like to verify the performance of the approach as well on large sources.

guybedford:
Ok the performance seems to be about 1ms per 800KiB of source.

That seems ok for now but I think it can likely be improved too by using some `lastIndex` optimizations.

Will merge for now though!

guybedford:
I just tested these sorts of cases again and I believe this bug has since been fixed in Firefox - so I may remove this change in a future release.

PR #38, opened 2019-08-21
Upgrade to latest Wasm es-module-lexer

guybedford:
This upgrades to the latest es-module-lexer, which now uses Web Assembly for the analysis. We still retain a single-file .js build at the cost of an extra 2KiB, but the win is 6x improvement in cold start performance, which improves the performance from over 100ms for multiple megabytes of JS analysis to down to under 50ms!

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,6 @@ Due to the use of a tiny [Web Assembly JS tokenizer for ES module syntax only](h
 
 Works in all browsers with [baseline ES module support](https://caniuse.com/#feat=es6-module).
 
-For support in Microsoft Edge, a [TextEncoder polyfill](https://github.com/anonyco/FastestSmallestTextEncoderDecoder) needs to be included in the environment.
-
 ### Import Maps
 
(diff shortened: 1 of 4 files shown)


daKmoR:
Won't that mean that browsers not supporting wasm can no longer be shimmed? looking at IE11...

guybedford:
@daKmoR this project specifically builds on native ES module support as its baseline, all of which browsers support Web Assembly. SystemJS is the project to use for IE11 support.

daKmoR:
ah I see - thx for clarifying 👍  totally makes sense 🤗

guybedford:
It looks like this is showing a roughly 30% overall performance improvement loading the full Angular build.

LarsDenBakker:
Perhaps it would be useful to also ship a version with the `TextEncoder` polyfill included? Just so that it's easy for people to just include one script tag and be on their way.

guybedford:
With https://github.com/guybedford/es-module-lexer/pull/15 merged, the TextEncoder dependency is no longer necessary now.

PR #39, opened 2019-08-21
Optimize sourcemap detection

guybedford:
This provides a lastIndex optimization for the regex approach. The remaining `new URL` call is still quite expensive and would benefit from using the internal resolver too.

If the URL implementation can be fixed, that gets the time down to about 0.1ms per 500KiB over 1ms per 500KiB, but without that the improvement to ~0.7ms doesn't yet seem worthwhile.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -114,7 +114,7 @@ export function resolveIfNotPlainOrUrl (relUrl, parentUrl) {
  *
  */
 
-function resolveUrl (relUrl, parentUrl) {
+export function resolveUrl (relUrl, parentUrl) {
   return resolveIfNotPlainOrUrl(relUrl, parentUrl) ||
       relUrl.indexOf(':') !== -1 && relUrl ||
       resolveIfNotPlainOrUrl('./' + relUrl, parentUrl);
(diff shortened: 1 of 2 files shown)


guybedford:
I've updated this approach to use the fast internal resolver over `new URL`.

@MicahZoltu is there any chance you could test out this PR against your app to verify it behaves as expected? This would be useful before releasing this feature to ensure the performance.

MicahZoltu:
@guybedford Any chance you could update the files in `dist` in this PR?  I would assume that has to be done anyway before this can be merged, and I don't currently have a dev environment setup for this project so generating the `dist` files for testing is more than the 2 clicks I was hoping for...

MicahZoltu:
I was able to test by copying the changes directly into the rolled up `dist/es-module-shim.js` and it appears to work (Firefox didn't give errors about being unable to find the source maps).  Would be good to test with a generated distributable, rather than a hacked one, but it seemed to work.

FWIW, my modules are tiny, so this is certainly not a stress test of any kind.  :)

guybedford:
Thanks so much for taking the time to try it out. Because the check runs against every source for this project that's more where the performance concern comes from. Will merge now and the release should follow in the next few days.

PR #55, opened 2019-08-30
Add test for execution order

LarsDenBakker:
On Edge modules are not executed in the original order they are imported with. This is a problem when code needs to rely on the side-effects of a module.

I added a test which fails randomly on Edge. I spent some time debugging this, I couldn't see a clear cause why it's not executed in the right order.

### DIFF
--- a/test/browser-modules.js
+++ b/test/browser-modules.js
@@ -49,7 +49,7 @@ suite('Basic loading tests', () => {
   test('should import json', async function () {
     var m = await importShim('./fixtures/json.json');
     assert.equal(m.default.json, 'module');
-  }); 
+  });
 
   test('should throw json parse errors', async function () {
     try {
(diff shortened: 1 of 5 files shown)


guybedford:
Can you confirm exactly which version of Edge you were testing on?

Would be very interesting if this turns out to be an Edge bug! Did you try the same case with normal modules and without ES module shims?

The old Edge is actually being maintained though so if it is an Edge bug we should move it there and document it in the readme.

If there are ways ES module shims could fix the bug, we should explore that too.

LarsDenBakker:
This is on Edge 17, but it happens on Edge 18 as well.

Much to my surprise, it's happening without es-module-shims as well, although the execution order is always reversed rather than random.

This loads fine:
```
    <script type="module" src="./fixtures/es-modules/execution-order-a.js"></script>
    <script type="module" src="./fixtures/es-modules/execution-order-b.js"></script>
    <script type="module" src="./fixtures/es-modules/execution-order-c.js"></script>
```

This loads in the reverse order:
```
    <script type="module">
      import "./fixtures/es-modules/execution-order-a.js";
      import "./fixtures/es-modules/execution-order-b.js";
      import "./fixtures/es-modules/execution-order-c.js";
    </script>
```

tpluscode:
I git this on Edge 18.18362.0 on Browserstack and locally it says 

Microsoft Edge 42.17134.1.0 
Microsoft EdgeHTML 17.17134

🤔

guybedford:
I've posted a ChakraCore bug at https://github.com/microsoft/ChakraCore/issues/6261.

guybedford:
Thanks for the test case, I've moved one manually into the fix PR at #57.

guybedford:
Fix released in 0.4.2.

Issue #54, opened 2019-08-30
application/x-javascript throws error


costingeana:
This may be related to #52. Some servers may also return content-type 'application/x-javascript' or 'text/x-javascript'; these values are specified in the HTML standard: https://mimesniff.spec.whatwg.org/#javascript-mime-type.

Should we support these values?

guybedford:
Are you specifically using servers with this support that it would be useful for? Yes in theory we should support all the MIMEs, but we should also aim for practicality, where the important point is covering supporting all the servers.

guybedford:
Released in 0.4.2.

PR #56, opened 2019-08-31
Support x-javascript MIME

guybedford:
Fixes https://github.com/guybedford/es-module-shims/issues/54.

@costingeana I would still value your input on what servers you found were producing this content type.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -173,50 +173,45 @@ function getOrCreateLoad (url, source) {
       load.r = res.url;
 
       const contentType = res.headers.get('content-type');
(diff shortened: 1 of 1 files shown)


costingeana:
@guybedford, after some investigation I found out that there are servers running `nginx` with this mime types configuration:
```
application/x-javascript           .js
```
In my opinion this piece of configuration has to be changed on each server because we can say that `application/x-javascript` is 'deprecated'. I can do that in my organization; however, this cannot be imposed to other organizations/people.

P.S. Sorry for my tardy answer. I was away for a couple of days.

guybedford:
Thanks @costingeana for following-up here, good to know there is good reason to support it.

PR #71, opened 2019-10-08
sourceMappingURL replacement may be easily broken by extra line breaks

just-dont:
...this improves detection. Admittedly, this doesn't exactly meet source map standard (which claims that `//# sourceMappingURL` must be "at the end of the file", but at least this meets browsers' behavior (checked with Firefox and Chrome) - they have no problems attempting to load source map even if mapping is in the middle of the file.

The basic idea is to pick last `//# sourceMappingURL=` from the file and then treat anything after `=` and until line feed or end of file as URL. This last part matches browsers' behavior as well.

Based on last discussion in this issue: https://github.com/guybedford/es-module-shims/issues/4

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -114,8 +114,8 @@ function resolveDeps (load, seen) {
     resolvedSource += source.slice(lastIndex);
   }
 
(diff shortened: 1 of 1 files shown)


guybedford:
This seems sensible as it leaves it up to the specific browser implementation how it is eventually handled exactly. Thanks for the PR.

PR #73, opened 2019-10-20
Implement fetch hook

guybedford:
This implements a very simple fetch hook for this project (two line change!).

It should open up some quite powerful workflows like TypeScript in-browser on native modules.


README.md
```diff
@@ -158,6 +160,36 @@ This matches the specification for ES module workers, supporting all features of
...
+
+For example:
+
+```js
+importShim.fetch = async function (url) {
+  const response = await fetch(url);
+  if (response.url.endsWith('.ts')) {
+    const source = await response.body();
+    const transformed = tsCompile(source);
+    return new Response(URL.createObjectURL(new Blob([transformed], { type: 'application/javascript' })), { status: 200 });
```
zaygraveyard: `URL.createObjectURL` is not needed (it returns a URL which would be used as contents of the response)
And the `status` is `200` by default, so it doesn't have to be explicitly specified

```js
return new Response(new Blob([transformed], { type: 'application/javascript' });
```

zaygraveyard:
This is awesome, it works like a charm, thank you :+1: 

Even-though I like and am in support of this feature, one nitpick I have is that up until this change, this library was more like a polyfill (it didn't do anything that wouldn't have been possible natively) but this feature is something that wont be possible natively. 
Maybe add a warning next to this feature in the README?

PS: My current use-case is only in development (for running tests)

PR #81, opened 2020-04-20
Gotta keep the stats up to date!

MicahZoltu:
If you reject this PR I would not be offended.  Just was reading the readme and noticed the out of date number at the top, figured I would PR a quick-fix.


guybedford:
Thanks, definitely worth doing!

Issue #94, opened 2020-10-01
Support a resolve hook


gingur:
Was wondering if it would be worth moving importMap support to a plugin, so it's considered optional? I have some ideas that don't align with the current importMap spec, and so want to replace it entirely with a different solution but hoping to ship w/out the extra bytes.

guybedford:
Are you sure the problem here is the file size impact of having import maps included?

The code for import maps in this project is not even 100 lines, minified it is really not a bloat problem.

Perhaps the focus here rather should be on ensuring the resolver hook can be properly exposed to add customizations?

gingur:
yeah, it depends on how robust you want to make the interface. in its current state, I'm unable to do much for my use cases, but moving importMap to a plugin would most likely open up things and expose critical utils needed.

for what it's worth, I was able to get a node implementation working, as well introduce some interesting preprocessing optimizations around the importMap definition by using your source with minimal changes. my hope would be to apply similar functionality as a plugin, instead of having to completely fork the project.

guybedford:
I would be happy to take a PR to this project that renames the internal `resolve` calls to instead call a `importShim.resolve` function which is the import maps etc implementation by default. That way this could then be hooked with custom changes more easily.

Happy to discuss the outline of the above further if that interests you.

If you'd consider wanting to upstream the Node.js support that is certainly something I'd be open to taking a PR for here as well.

gingur:
Sure, I can take a crack at it.

guybedford:
Great, let's start with the resolve hook, and if you're also interested in the node implementation to create a new issue to outline the approach there.

guybedford:
This was implemented in https://github.com/guybedford/es-module-shims/pull/146.

PR #95, opened 2020-10-07
Expose resolve and resolveImportMap and global importShimOptions

gingur:
https://github.com/guybedford/es-module-shims/issues/94

Few things this introduces
 * Enables a global `importShimOptions` object to be defined before importShim is loaded, this allows the consumer to specify overrides before the "firstTopLevelProcess" is called.
 * Exposes `resolve` and `resolveImportMap` functions that can now be overridden via `importShimOptions` or by directly mutating `importShim`
 * Exposes a `defaults` object that overridden functions could leverage to use default behavior
 * Add `importMap` argument to all `importShim.resolve` calls
 * Added a `start` script to run a simple http server to run the `test` HTML files in the default browser
 * Removed unused dependencies

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -2,7 +2,14 @@ import { baseUrl as pageBaseUrl, resolveImportMap, createBlob, resolveUrl, resol
 import { init, parse } from '../node_modules/es-module-lexer/dist/lexer.js';
 
 let id = 0;
+let waitingForImportMapsInterval;
+let firstTopLevelProcess = true;
+let lastLoad;
+let importMap = { imports: {}, scopes: {}, depcache: {} };
+let importMapPromise = resolvedPromise;
+const meta = {};
 const registry = {};
+const edge = navigator.userAgent.match(/Edge\/\d\d\.\d+$/);
 
 async function loadAll (load, seen) {
   if (load.b || seen[load.u])
@@ -12,8 +19,6 @@ async function loadAll (load, seen) {
   return Promise.all(load.d.map(dep => loadAll(dep, seen)));
 }
 
-let waitingForImportMapsInterval;
-let firstTopLevelProcess = true;
 async function topLevelLoad (url, source) {
   if (waitingForImportMapsInterval > 0) {
     clearTimeout(waitingForImportMapsInterval);
@@ -21,7 +26,7 @@ async function topLevelLoad (url, source) {
   }
   if (firstTopLevelProcess) {
     firstTopLevelProcess = false;
-    processScripts();
+    importShim.load();
   }
   await importMapPromise;
   await init;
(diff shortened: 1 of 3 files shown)


src/es-module-shims.js
```diff
@@ -229,17 +209,39 @@ async function processScripts () {
...
 function throwUnresolved (id, parentUrl) {
   throw Error("Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'"));
 }
+
+const importShimOptions = self.importShimOptions || {};
+const importShimDefaults = {
+  fetch: url => fetch(url),
+  skip: /^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//,
+  load: processScripts,
+  resolveImport: resolve,
```
guybedford, line 232: Let's just call this `resolve`.

src/es-module-shims.js
```diff
@@ -229,17 +209,39 @@ async function processScripts () {
...
-function resolve (id, parentUrl) {
+function resolve (id, parentUrl, importMap) {
   return resolveImportMap(importMap, resolveIfNotPlainOrUrl(id, parentUrl) || id, parentUrl) || throwUnresolved(id, parentUrl);
 }
 
 function throwUnresolved (id, parentUrl) {
   throw Error("Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'"));
 }
+
+const importShimOptions = self.importShimOptions || {};
```
guybedford, line 227: I would prefer not using globals in an API. Is it absolutely necessary to your use case?

src/es-module-shims.js
```diff
@@ -229,17 +209,39 @@ async function processScripts () {
...
+const importShimDefaults = {
+  fetch: url => fetch(url),
+  skip: /^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//,
+  load: processScripts,
+  resolveImport: resolve,
+  resolveImportMap: resolveAndComposeImportMap,
```
guybedford, line 233: Rather than making this hook available, it could be nice to just handle this by hooking `load` entirely, and re-exposing the `importMap` property here for custom changes.

guybedford (commented):
Thanks for the PR!

src/es-module-shims.js
gingur: the current default behavior immediately calls `importShim.load()`, giving no way to customize the shim before it starts processing module/importmap shims on the DOM. an alternative would be to introduce a breaking change and remove the immediate `load` call in favor of the consumer having to manually call it.

src/es-module-shims.js
gingur: the `load` hook IMO doesn't have much value, as there is too much closure based logic going on to make it reliably possible for a consumer to override. the intent of `resolveImportMap` is to simply hook into the logic of translating the json of a importmap-shim into a digestible format.  this will allow the json to be in a non-standard/compressed format but via the `resolveImportMap` be able to decompress into the format required.

src/es-module-shims.js
guybedford: Another alternative might be to run the initial `load` call on a zero timeout so that a sync script could still alter options before it starts.

src/es-module-shims.js
guybedford: Are you saying the problem here is that you still want the `module-shim` scripts to load, while overriding the `importmap-shim` scripts?

Perhaps we could split load into two functions - `loadMaps` and `loadScripts`. Where those two functions are the top-level hooks that run one after the other.

src/es-module-shims.js
gingur: I can give that a shot, as for overriding should there be `importShim.config` function to apply? or should the consumer just directly mutate the global?

src/es-module-shims.js
gingur: I'm looking for a hook that takes the data from `<script type="importmap-shim">` and then returns a `importMap` that `importShim` is familiar with (imports/scopes/depcache). This would allow the format of the data in `<script type="importmap-shim">` to be in *any* format, so long as the `importShim.resolveImportMap` mutates it to the expected shape.

```html
<script>
  importShim.resolveImportMap = (json, baseUrl, parentMap) => {
    // transform json to a new importMap, which would be the same shape as parentMap
    return newImportMap;
  };
</script>
<script type="importmap-shim" src="http://mycustom.importmap.format.json"></script>

src/es-module-shims.js
guybedford: Mutation is fine.

src/es-module-shims.js
guybedford: We must just have a single `loadMaps` function, it's only one line to do your own "search and fetch" for import maps.

guybedford:
This is out of date at this point. If you'd like to reconsider working on any of these features I'm always happy to discussion and review PRs further.

PR #96, opened 2020-10-10
onerror hook should rethrow error by default

w72:


guybedford:
Thanks for the PR, I'll post this out with the next release in the coming weeks.

PR #98, opened 2020-11-17
Remove firefox sourceMappingURL rebase

guybedford:
This removes https://github.com/guybedford/es-module-shims/pull/37 which I've tested in Firefox to no longer be necessary.

//cc @MicahZoltu


MicahZoltu:
Unfortunately I no longer remember what the problem was or how to reproduce it.  If you say that the problem is gone, I'll take your word for it.  😄

PR #102, opened 2020-12-03
Allow scripts to dynamically be added

### LINKED ISSUE #103: `importShim.load()` doesn't load new scripts or maps
### Description

Calling `importShim.load()` exits early and doesn't load dynamically added  `module-shim` or `importmap-shim` scripts.

ifiokjr:
### Description

Currently, if any script has been processed previously the `for` loop within `processScripts` function causes the function to return early. 

This is changed to `continue` to allow processing of new `module-shim` and `importmap-shim` scripts when calling `importShim.load()`.

Fixes #103


guybedford:
Woah good catch :) Thanks for the PR.

PR #104, opened 2020-12-03
Add TypeScript declaration file

ifiokjr:


guybedford:
Nice. It's wordy, but I assume most editors format these types well if you're suggesting the pattern here?

ifiokjr:
> It's wordy
> @guybedford

True 😄 

I guess it's a personal preference I have for inline documentation which can save a few keystrokes by preventing the need to open the docs. 

Here's how it appears to users in **VSCode**.

guybedford:
Yeah I like the style.

PR #107, opened 2020-12-23
Support URL to URL mappings

### LINKED ISSUE #97: Support URL to URL mappings
URL to URL mappings are currently supported in the import map - per the spec support for this should be added.

vovacodes:
Fixes #97

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -151,9 +151,9 @@ export function resolveUrl (relUrl, parentUrl) {
   return resolveIfNotPlainOrUrl(relUrl, parentUrl) || (relUrl.indexOf(':') !== -1 ? relUrl : resolveIfNotPlainOrUrl('./' + relUrl, parentUrl));
 }
 
-function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap) {
+function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap, scopeUrl) {
   for (let p in packages) {
-    const resolvedLhs = resolveIfNotPlainOrUrl(p, baseUrl) || p;
+    const resolvedLhs = resolveIfNotPlainOrUrl(p, scopeUrl || baseUrl) || p;
     let target = packages[p];
     if (typeof target !== 'string') 
       continue;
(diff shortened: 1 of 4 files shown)


src/common.js
```diff
@@ -151,13 +151,13 @@ export function resolveUrl (relUrl, parentUrl) {
...
-function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap) {
+function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap, parentUrl) {
   for (let p in packages) {
-    const resolvedLhs = resolveIfNotPlainOrUrl(p, baseUrl) || p;
+    const resolvedLhs = resolveIfNotPlainOrUrl(p, parentUrl != undefined ? parentUrl : baseUrl) || p;
     let target = packages[p];
     if (typeof target !== 'string') 
       continue;
-    const mapped = resolveImportMap(parentMap, resolveIfNotPlainOrUrl(target, baseUrl) || target, baseUrl);
+    const mapped = resolveImportMap(parentMap, resolveIfNotPlainOrUrl(target, baseUrl) || target, parentUrl);
```
guybedford, line 160: The target itself must always be resolved relative to the `baseUrl` not the scope Url.

src/common.js
```diff
@@ -151,13 +151,13 @@ export function resolveUrl (relUrl, parentUrl) {
   return resolveIfNotPlainOrUrl(relUrl, parentUrl) || (relUrl.indexOf(':') !== -1 ? relUrl : resolveIfNotPlainOrUrl('./' + relUrl, parentUrl));
 }
 
-function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap) {
+function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap, parentUrl) {
```
guybedford, line 154: Can we call this `scopeUrl`?

src/common.js
```diff
@@ -151,13 +151,13 @@ export function resolveUrl (relUrl, parentUrl) {
...
   for (let p in packages) {
-    const resolvedLhs = resolveIfNotPlainOrUrl(p, baseUrl) || p;
+    const resolvedLhs = resolveIfNotPlainOrUrl(p, parentUrl != undefined ? parentUrl : baseUrl) || p;
```
guybedford, line 156: If we are happy with `!=` semantics then `scopeUrl || baseUrl` is also fine :)

guybedford (approved):
Perfect, that's the fix - thanks!

src/common.js
vovacodes: `scopeUrl || baseUrl` wouldn't will return `baseURL` id `scopeURL` is an empty string, I'm not sure if an empty string can occur in that position though

guybedford: Good point. Yes best not to fret over these details so happy to leave as-is. An empty string is permitted, but the assumption in this code here is that the URLs are already normalized so should never be the empty string I believe.

Issue #108, opened 2020-12-24
Fetch hook doesn't always get applied before initial load


lewisl9029:
I recently ran into a bug where `importShim.load()` would very occasionally get called before a subsequent script has the chance to set the fetch script in a new project I'm working on.

I've observed that it generally happens when the subsequent script ends up taking a significant amount of time to load compared to es-module-shims: 
![image](https://user-images.githubusercontent.com/6934200/103046326-b14e8700-453c-11eb-8c10-81a2e97d763b.png)

Here's what my script loading order looks like: 

```
  <body>
    <link href="/index.css" rel="stylesheet" />
    <noscript>You need to enable JavaScript to run this app.</noscript>
    <div id="root"></div>
    <script type="module" src="/es-module-shims.js"></script>
    <script type="module" src="/dev.js"></script>
    <script type="importmap-shim" src="/production.importmap"></script>
    <script type="module-shim" src="/index.js"></script>
  </body>
```

I've forked the jspm.dev bundle of es-module-shims to a local file and added a log statement (`processing scripts
`) at https://github.com/guybedford/es-module-shims/blob/master/src/es-module-shims.js#L218

And added a log at the beginning of the dev.js file that sets the fetch hook (`setting fetch shim`).

This is what the logs look like when the error occurs:
![image](https://user-images.githubusercontent.com/6934200/103046477-4a7d9d80-453d-11eb-869b-f0889e657650.png)

This is what the logs normally look like:
![image](https://user-images.githubusercontent.com/6934200/103046559-8ca6df00-453d-11eb-80c7-4e39e1be70f0.png)

Note how the order between `setting fetch shim` and `processing scripts` is reversed.

I also tried to change https://github.com/guybedford/es-module-shims/blob/master/src/es-module-shims.js#L213 to `setTimeout(processScripts, 0)` as per this discussion: https://github.com/guybedford/es-module-shims/pull/95/files#r500766794

But it didn't seem to have much of an effect on the frequency of occurrence, possibly because the loading of dev.js took longer than the next async tick.

Would it make sense to provide some way for users to provide the fetch hook before loading es-module-shims (i.e. so I can swap the loading order of dev.js and es-module-shims here)? Happy to try out any other solutions folks might have in mind as well.

guybedford:
In the example above, is your fetch hook defined in the `/dev.js` file?

[...]

PR #109, opened 2021-01-02
Add initial options API

lewisl9029:
This PR is meant to address the issue brought up in https://github.com/guybedford/es-module-shims/issues/108, whereby the fetch hook doesn't always get applied in time for the first shimmed module load.

We now accept an `importShimInitialOptions` global to apply specific options on initialization instead of the defaults, which can be supplied by the user before es-module-shims is loaded, guaranteeing that they will be applied to the first shimmed module load.

The API `importShimInitialOptions` was inspired by https://github.com/guybedford/es-module-shims/pull/95, but with the `Initial` qualifier added to hopefully avoid misuse since setting it after es-module-shims is loaded doesn't have any effect. The implementation also differs in that it explicitly applies only specific properties from the global object instead of merging onto it, to keep the API surface minimal and remove the possibility of shadowing other properties that the implementation depends on.

I've also added a new test suite for this. Not sure if the implementation is ideal, happy to take suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,26 +163,57 @@ then allowing dynamic injection of `<script type="importmap-shim">` to immediate
 
 This follows the [dynamic import map specification approach outlined in import map extensions](https://github.com/guybedford/import-maps-extensions).
 
-### Skip Processing
+### Init Options
 
 > Stability: Non-spec feature
+
+Provide a `esmsInitOptions` on the global scope before `es-module-shims` is loaded to configure various aspects of the module loading process:
+
+```html
+<script>
+  globalThis.esmsInitOptions = {
+    fetch: (url => fetch(url)),
+    skip: /^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//,
+    onerror: ((e) => { throw e; }),
+  }
+</script>
+<script defer src="es-module-shims.js"></script>
+```
+
+See below for a detailed description of each of these options.
+
+#### Skip Processing
 
 When loading modules that you know will only use baseline modules features, it is possible to set a rule to explicitly
 opt-out modules from rewriting. This improves performance because those modules then do not need to be processed or transformed at all, so that only local application code is handled and not library code.
 
-This can be configured by setting the `skip` URL regular expression on a `esmsInitOptions` global before loading es-module-shims:
+This can be configured by providing a URL regular expression for the `skip` option:
 
 ```js
-globalThis.esmsInitOptions = {
-  skip: /^https:\/\/cdn\.com/
-}
+<script>
+  globalThis.esmsInitOptions = {
+    skip: /^https:\/\/cdn\.com/ // defaults to `/^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//`
+  }
+</script>
+<script defer src="es-module-shims.js"></script>
 ```
 
 By default, this expression supports `jspm.dev`, `dev.jspm.io` and `cdn.pika.dev`.
 
-### Fetch Hook
+#### Error hook
 
-> Stability: Non-spec feature
+You can provide a function to handle errors during the module loading process by providing a `onerror` option:
+
+```js
+<script>
+  globalThis.esmsInitOptions = {
+    onerror: error => console.log(error) // defaults to `((e) => { throw e; })`
+  }
+</script>
+<script defer src="es-module-shims.js"></script>
+```
+
+#### Fetch Hook
 
 This is provided as a convenience feature since the pipeline handles the same data URL rewriting and circular handling of the module graph that applies when trying to implement any module transform system.
 
@@ -191,17 +222,20 @@ The ES Module Shims fetch hook can be used to implement transform plugins.
 For example:
 
 ```js
-globalThis.esmsInitOptions = {
-  fetch: async function (url) {
-    const response = await fetch(url);
-    if (response.url.endsWith('.ts')) {
-      const source = await response.body();
-      const transformed = tsCompile(source);
-      return new Response(new Blob([transformed], { type: 'application/javascript' }));
-    }
-    return response;
+<script>
+  globalThis.esmsInitOptions = {
+    fetch: async function (url) {
+      const response = await fetch(url);
+      if (response.url.endsWith('.ts')) {
+        const source = await response.body();
+        const transformed = tsCompile(source);
+        return new Response(new Blob([transformed], { type: 'application/javascript' }));
+      }
+      return response;
+    } // defaults to `(url => fetch(url))`
   }
-}
+</script>
+<script defer src="es-module-shims.js"></script>
 ```
 
 Because the dependency analysis applies by ES Module Shims takes care of ensuring all dependencies run through the same fetch hook,
@@ -238,7 +272,7 @@ globalThis.esmsInitOptions = {
 }
 ```
 
-#### Plugins
+##### Plugins
 
 Since the Fetch Hook is very new, there are no plugin examples of it yet, but it should be easy to support various workflows
 such as TypeScript and new JS features this way.
--- a/README.md
+++ b/README.md
@@ -170,10 +170,10 @@ This follows the [dynamic import map specification approach outlined in import m
 When loading modules that you know will only use baseline modules features, it is possible to set a rule to explicitly
 opt-out modules from rewriting. This improves performance because those modules then do not need to be processed or transformed at all, so that only local application code is handled and not library code.
 
-This can be configured by setting the `skip` URL regular expression on a `importShimInitialOptions` global before loading es-module-shims:
+This can be configured by setting the `skip` URL regular expression on a `esmsInitOptions` global before loading es-module-shims:
 
 ```js
-self.importShimInitialOptions = {
+self.esmsInitOptions = {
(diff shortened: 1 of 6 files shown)


src/es-module-shims.js
```diff
@@ -57,12 +57,11 @@ Object.defineProperties(importShim, {
   l: { value: undefined, writable: true },
   e: { value: undefined, writable: true }
 });
-importShim.fetch = url => fetch(url);
-importShim.skip = /^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//;
+const importShimInitialOptions = self.importShimInitialOptions || {};
+importShim.fetch = importShimInitialOptions.fetch || (url => fetch(url));
+importShim.skip = importShimInitialOptions.skip || /^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//;
+importShim.onerror = importShimInitialOptions.onerror || ((e) => { throw e; });
```
guybedford, line 63: In fact, we don't even need to expose these as instance-level hooks anymore, perhaps we make this a major change and treat `importShim` as fully stateless now.

src/es-module-shims.js
```diff
@@ -57,12 +57,11 @@ Object.defineProperties(importShim, {
...
 importShim.load = processScripts;
```
guybedford, line 64: Then maybe we could invert the control here and have a `init` hook that returns a bag with the "instance object" (no longer `importShim` itself) including the `load` function.

README.md
```diff
@@ -178,6 +163,29 @@ then allowing dynamic injection of `<script type="importmap-shim">` to immediate
...
+
+> Stability: Non-spec feature
+
+When loading modules that you know will only use baseline modules features, it is possible to set a rule to explicitly
+opt-out modules from rewriting. This improves performance because those modules then do not need to be processed or transformed at all, so that only local application code is handled and not library code.
+
+This can be configured by setting the `skip` URL regular expression on a `importShimInitialOptions` global before loading es-module-shims:
+
+```js
+self.importShimInitialOptions = {
```
guybedford, line 176: How about calling it `esmShimsInitOptions = {...}`?

Or maybe `esmsInitOptions` or `ESMShimsInitOptions`?

guybedford (approved):
This looks great, thank you for working on this PR @lewisl9029.

Apologies for the delayed response, only just got back from holiday here.

I think we could lean in more to this model and actually deprecate the former hooks system. Let me know your thoughts.

src/es-module-shims.js
lewisl9029: One reason I didn't remove the original API is this new API doesn't support changing the hook dynamically after the initial load by overwriting the hooks on the `importShim` instance. I wasn't sure if there might be valid use cases in the wild for the ability to swap out these hooks for subsequent module loads. 

If you're not aware of any use cases either, and don't mind removing support for it, I'd be more than happy to remove the old API. 🙂 

I usually get pretty nervous with these kinds of directly mutative APIs since they open up the possibility for some really nasty edge cases when users make mutations at inconvenient times. If we wanted to eventually re-introduce the ability to update these hooks after the initial load, I'd probably lean towards introducing some update function for users to call explicitly to update the hooks internally. That way we'd at least know exactly what's getting updated and when, so we can prepare the internals to accept the updates before actually applying them.

src/es-module-shims.js
lewisl9029: Wanted to confirm we're on the same page here since I'm not sure if I've fully understood your comment. 

Did you mean asking the user to supply a init function that returns the options object instead of supplying an options object directly?

```js
// in some user supplied script before esms loads
self.esmsInit = () => {
  return {
    // whatever the user wants to supply for options
  }
}

// in esms implementation
const userSuppliedOptions = self.esmsInit ? self.esmsInit() : {}

const options = {
  fetch: userSuppliedOptions.fetch || (url => fetch(url)),
  skip: userSuppliedOptions.skip || /^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//,
  onerror: userSuppliedOptions.onerror || ((e) => { throw e; }),
  load: userSuppliedOptions.load || processScripts
}

// ...

if (hasDocument) {
  options.load();
  waitingForImportMapsInterval = setInterval(options.load, 20);
}
```

Curious what the benefit would be over just asking for an object? Or maybe I'm completely off base with my interpretation of your comment? Please clarify if that's the case. Or if you feel it might be easier, feel free to push changes onto this PR directly.

guybedford: I was thinking something like:

```js
globalThis.esmsInit = {
  fetch: fetchHook,
  skip,
  onerror,
  init ({ load, ...any other calls / info / useful runtime instance state }) {
    bindLoadCall(load);
  }
};

```

lewisl9029: It sounds like you probably have some use case in mind for this init hook that I might be still missing.

I've pushed an update with just the suggested rename for now: https://github.com/guybedford/es-module-shims/pull/109/commits/06410f9567da40b76e418e2f4072ccef07fc033d

It might be easier at this point to get this merged in as is, so you can follow up with all the changes you had in mind for the next release without any further back and forth to get me up to speed. Wdyt?

guybedford: The `importShim` instance was always a hack, so the main suggestion was just to treat it as a function only and remove all state from that global which seems like a nice simplification your work enables.

Certainly we can treat that as a separate PR though.

README.md
```diff
@@ -178,6 +163,29 @@ then allowing dynamic injection of `<script type="importmap-shim">` to immediate
...
+
+> Stability: Non-spec feature
+
+When loading modules that you know will only use baseline modules features, it is possible to set a rule to explicitly
+opt-out modules from rewriting. This improves performance because those modules then do not need to be processed or transformed at all, so that only local application code is handled and not library code.
+
+This can be configured by setting the `skip` URL regular expression on a `esmsInitOptions` global before loading es-module-shims:
+
+```js
+self.esmsInitOptions = {
```
guybedford, line 176: To make this clearer can we instead use `globalThis` or `window`. Self might be confusing to some users.

guybedford: It would also be nice to have a full example of how to load both es-module-shims and set the options together like a user would in practice.

guybedford: Alternatively even a dedicated "init options" section to discuss this mechanic explicitly.

src/es-module-shims.js
lewisl9029: Here's my attempt at decoupling all the state from the importShim function: https://github.com/guybedford/es-module-shims/pull/109/commits/780a1f7ab53a7240aa60270d22ee5c3ca5cadaca

Basically I've moved everything that used to be on `importShim` to a separate `esmsState` object. Initially I tried to not expose it globally, but it seemed like this import.meta implementation required it: https://github.com/guybedford/es-module-shims/blob/master/src/es-module-shims.js#L122

Definitely open to other suggestions on ways to work around that.

README.md
lewisl9029: I've added the Init Options section here: https://github.com/guybedford/es-module-shims/pull/109/commits/8b3d26ddc7771610454be0ad8d2951129b531dde

And added a small blurb to describe the undocumented onerror hook.

Let me know if there's anything else you'd like to see.

src/es-module-shims.js
```diff
@@ -52,17 +52,20 @@ async function importMetaResolve (id, parentUrl = this.url) {
   return resolve(id, `${parentUrl}`);
 }
 
-Object.defineProperties(importShim, {
+let esmsState = {}
+self.esmsState = esmsState
```
guybedford, line 56: Of course this has to be exposed, sorry I overlooked that here.

Sorry to be difficult but I think we should probably just leave it as `importShim` then if we need a global anyway since that's to hand. Alternatively a private name like `_esmsState`.

src/es-module-shims.js
```diff
@@ -52,17 +52,20 @@ async function importMetaResolve (id, parentUrl = this.url) {
...
   e: { value: undefined, writable: true }
 });
-importShim.fetch = url => fetch(url);
-importShim.skip = /^https?:\/\/(cdn\.pika\.dev|dev\.jspm\.io|jspm\.dev)\//;
-importShim.load = processScripts;
-importShim.onerror = (e) => {
-  throw e;
-};
+
+const esmsInitOptions = self.esmsInitOptions || {};
```
guybedford, line 64: Should we delete this global options object after using it?

guybedford (approved):
Thanks again for pushing this through, sorry for delay been slowly getting back into things.

This looks almost ready to go to me, just a couple more questions about the global handling.

src/es-module-shims.js
lewisl9029: Understood, `_esmsState` sounds like the better option to me since it at least allows us to separate the public API from the internal state, even if it's still technically accessible on the global scope.

Addressed here: https://github.com/guybedford/es-module-shims/pull/109/commits/348e65b2f902afeb0e2361c421276e72247a45d5

src/es-module-shims.js
lewisl9029: Addressed here: https://github.com/guybedford/es-module-shims/pull/109/commits/b2e8e0b2422b3c77cee8cccadc6e1994d936165d

PR #123, opened 2021-05-14
Fix immediately added dynamic import maps

vovacodes:
This draft adds a failing test (`http://localhost:8080/test/dynamic-import-map-shim.html`) for the scenario when an import map is inserted dynamically from JS and `importShim()` that is trying to import a bare module specifier defined in that dynamic import map is called right after the insertion.

In the state after the [first commit](https://github.com/guybedford/es-module-shims/pull/123/commits/6504a33cab7c9e8cd2e153249a4f67ca18dcd867) the code works in Safari and prints the following to the console

image

but it fails in Chrome with printing the following:

image

The second commit fixes it.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -64,9 +64,7 @@ async function topLevelLoad (url, source, polyfill) {
 }
 
 async function importShim (id, parentUrl = pageBaseUrl) {
-  // Give dynamically inserted import maps a chance to be processed
-  // if `importShim()` was called synchronously after the insertion.
-  await new Promise(resolve => setTimeout(resolve, 0))
+  await featureDetectionPromise;
   // Make sure all the "in-flight" import maps are loaded and applied.
   await importMapPromise;
   return topLevelLoad(resolve(id, parentUrl).r || throwUnresolved(id, parentUrl));
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -64,6 +64,9 @@ async function topLevelLoad (url, source, polyfill) {
 }
 
 async function importShim (id, parentUrl = pageBaseUrl) {
+  console.log('importShim before await', id)
+  await importMapPromise;
+  console.log('importShim after await', id)
   return topLevelLoad(resolve(id, parentUrl).r || throwUnresolved(id, parentUrl));
 }
 
@@ -291,6 +294,7 @@ new MutationObserver(mutations => {
     if (mutation.type !== 'childList') continue;
     for (const node of mutation.addedNodes) {
       if (node.tagName === 'SCRIPT' && node.type)
+        console.log('MutationObserver', node.innerHTML)
         processScript(node, !firstTopLevelProcess);
     }
   }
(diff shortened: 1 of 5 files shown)


src/es-module-shims.js
```diff
@@ -64,6 +64,9 @@ async function topLevelLoad (url, source, polyfill) {
 }
 
 async function importShim (id, parentUrl = pageBaseUrl) {
+  console.log('importShim before await', id)
+  await importMapPromise;
```
vovacodes, line 68: Without this line even the main inline import map isn't yet processed in Safari when the first `importShim` is called.

guybedford: Yes that is what this line is for - to wait on the currently loading import map.

guybedford: Did you verify if adding:

```suggestion
  await Promise.all([importMapPromise, new Promise(resolve => setTimeout(resolve, 0))]);
```

solved the issue here?

guybedford (commented):
Great to have a test. Let me know if I can help look into it further? I don't currently have a mac but could test on my iphone at least.

src/es-module-shims.js
vovacodes: no, let me try!

vovacodes: ah, it's probably not gonna work because the problem is until `new Promise(resolve => setTimeout(resolve, 0))` is resolved `importMapPromise` will be a reference to the "old" promise, that isn't yet chained with the "dynamic import map" promise

vovacodes:
@guybedford I think I'm ready to merge it if my response to your comments makes sense to you.

src/es-module-shims.js
```diff
@@ -64,6 +64,11 @@ async function topLevelLoad (url, source, polyfill) {
 }
 
 async function importShim (id, parentUrl = pageBaseUrl) {
+  // Give dynamically inserted import maps a chance to be processed
+  // if `importShim()` was called synchronously after the insertion.
+  await new Promise(resolve => setTimeout(resolve, 0))
```
guybedford, line 69: If you replace this with `await featureDetectionPromise` does that work? I think that might avoid an issue for early `importShim` calls as well actually...

src/es-module-shims.js
guybedford: Makes sense!

src/es-module-shims.js
guybedford: I've verified that this works. Please confirm, otherwise we are good to merge.

guybedford (approved):

src/es-module-shims.js
vovacodes: @guybedford I was going to test if it'll work for the subsequent insert-imports, like after `featureDetectionPromise` is already resolved awaiting it won't introduce a task-long delay that might be necessary to make sure the mutation observer is triggered and the import map is processed, wdyt?

guybedford: Awaiting a promise always introduces a task delay. But yes we should verify it is enough of a delay too.

vovacodes: isn't it a microtask though?

vovacodes: anyways, I'll check that real-quick now

guybedford: Ah I see what you mean, ok we may well need to retain the previous. I think featureDetectionPromise should be in here for a possible edge case though too.

vovacodes: hm it works in Chrome, Safari and Firefox for me but I'm indeed a bit worried there's still a race here that might lead to  some non-deterministic outcome. So yeah I'd prefer to bring back that task delay and will also keep the featureDetectionPromise

guybedford: I'd prefer not to be superstitious to be honest... if it works then it works!

vovacodes: you think? okay let's do that, we can always fix it later, we have a test now

guybedford: Thanks, yeah all codebases become collections of superstitions over time. Would be nice to avoid it unnecessarily :P

vovacodes:
@guybedford do you mind publishing a new version?

PR #124, opened 2021-05-15
Revoke blob URLs after the module is loaded

vovacodes:


src/es-module-shims.js
```diff
@@ -54,15 +54,34 @@ async function topLevelLoad (url, source, polyfill) {
...
   // if the top-level load is a shell, run its update function
-  if (load.s)
+  if (load.s) {
     (await dynamicImport(load.s)).u$_(module);
+  }
+  revokeObjectURLs(Object.keys(seen));
   return module;
 }
 
+function revokeObjectURLs(registryKeys) {
```
guybedford, line 71: Wondering if we should make this async and have an async tick eg every 100 revocations?

guybedford (approved):
I really like this scheduling technique. We could possibly even apply this to the ESM analysis itself to break up jank during the "instantiate" phase.

vovacodes:
@guybedford ha, good point actually, I can maybe look into that later 👍

PR #128, opened 2021-05-28
Fix dynamic imports in inline scripts

heypiotr:
Attempting to polyfill a dynamic import in an inline script currently fails with `TypeError: url is undefined`.

image

image

`load.r` is only set when the source needs to be fetched, which is not the case for inline scripts.

https://github.com/guybedford/es-module-shims/blob/986733009f4c9374decd4c0348740134ff499157/src/es-module-shims.js#L240-L244

My proposed solution is to rewrite such imports as `importShim("...", undefined)` … which seems to work, as it's gonna default to `pageBaseUrl`.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -184,7 +184,7 @@ function resolveDeps (load, seen) {
       }
       // dynamic import
       else {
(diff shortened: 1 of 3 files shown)


guybedford (approved):

guybedford:
Looks great, thanks for the fix. Will post out a release cycle in the next week.

Issue #129, opened 2021-06-04
Errors in FF and Safari (non-polyfilled module is imported?)


vovacodes:
I'm running into some weird errors with es-module-shims on this page (reproducible in FF and Safari): https://deploy-preview-51--romantic-wiles-3910cf.netlify.app/ 
I looks like the original (non-processed by the lexer) module is getting imported somewhere in the chain.

![image](https://user-images.githubusercontent.com/1524432/120767826-ddbef400-c51b-11eb-96c9-003d150a061b.png)

guybedford:
@vovacodes did you mean to be running in shim mode here? Which module is executing that you would not have expected to be executing?

vovacodes:
FocusManager.js is being evaluated it looks like. I expect it to run in the polyfill mode here

guybedford:
I see, agreed this is a bug in polyfill mode, looking into it.

guybedford:
@vovacodes is there perhaps a cycle in play with `FocusManager.js`? If so, that could be the reason the check algorithm is breaking down.

guybedford:
Do you think you could host a version of this running the debug build of ES module shims? That would help a lot to investigate how this is happening in the registry.

guybedford:
Fixed in #130.

PR #130, opened 2021-06-05
Fix polyfill checking logic

guybedford:
This fixes some of the polyfill logic, and should resolve the issue in https://github.com/guybedford/es-module-shims/issues/129.

The root of this was that we check if a module needs to be polyfilled or not based on detecting whether it can run natively. Then if a dependency requires polyfilling we must polyfill the parent importer as well. In this case, the logic of the dependency causing the parent to need polyfilling wasn't always carrying through, resulting in a native execution attempt which would fail.

At the same time I've updated shim mode to be turned out on by the presence of any shim script as documented in the readme (previously it was in a sort of mixed mode).

Finally I've made a debug build where setting `globalThis.ESMS_DEBUG = true` can allow registry inspection for debugging these kinds of complex registry issues, where the minified build removes this codepath.


vovacodes (approved):
Looks great! Let's ship it 👏🏼

guybedford:
Published as 0.10.6.

Issue #127, opened 2021-05-28
Feature detection hangs in Safari on file://


heypiotr:
Test file: [MyPrototype.html.zip](https://github.com/guybedford/es-module-shims/files/6560971/MyPrototype.html.zip)

Steps to repro: download the test file, and open it in Safari directly from the filesystem.

What happens? There's a CORS error when the importmap-detection iframe tries to call into the parent:

![image](https://user-images.githubusercontent.com/1155764/119987640-896cbf00-bfc5-11eb-8d5b-3936d822ea78.png)

… which means _$s never gets called, which means the feature detection promise never resolves.

Relevant code: https://github.com/guybedford/es-module-shims/blob/986733009f4c9374decd4c0348740134ff499157/src/common.js#L56-L67

Interestingly enough, this works fine in Firefox on file:// 🤷 I'm not entirely surely which behavior is "correct", should a null-origin blob:null:// iframe be able to call into the null-origin file:// parent, or not?

I experimented with using postMessage instead of trying to call into the parent directly, but even then I was getting "Unable to post message to null. Recipient has origin null."

I realize file:// support probably isn't high on the list, but I figured I'll share my findings anyway, maybe somebody will have some clever idea about this!

guybedford:
Since Safari currently has zero support for import maps, we could simply revert to a version sniffing detection for this case specifically. If you'd like to create a PR to implement that I'd be happy to blanket assume the feature outcome for Safari, we will be eagerly tracking support anyway!

guybedford:
(it would be fine to even make the default feature detection apply regardless of file URLs or not)

heypiotr:
(Sorry for late response, vacation 🌴)

Thanks for the suggestion, happy to work on a PR! When you say "version sniffing detection", do you mean just looking at the user agent string, similar to Edge?

https://github.com/guybedford/es-module-shims/blob/95e15c829389dd86168edc50a7858830bde5e49f/src/es-module-shims.js#L100

guybedford:
No problem, yes that's what I mean. Alternatively if there is a way to detect if we have origin policies blocking feature detection or a way to do feature detection without violating the policy that could work too.

PR #131, opened 2021-06-09
Fix supportsImportMaps feature detection on null origins

heypiotr:
Fixes https://github.com/guybedford/es-module-shims/issues/127.

In Safari (and Chrome\*), import maps feature detection via a `blob:` iframe fails on null origins (such as `file:`, but I guess it might also fail e.g. in sandboxed iframes) — `parent._$s(v)` fails with a CORS error.

_\* This of course is not really a problem in Chrome, since they support import maps natively anyway. Interestingly enough, it doesn't fail in Firefox, so it looks like Firefox has more lax null origin restrictions?_

This PR proposes to use [`srcdoc`](https://caniuse.com/?search=srcdoc) instead of `src="blob:...` for the iframe, which seems to work.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -62,7 +62,7 @@ export const featureDetectionPromise = Promise.all([
     };
     const iframe = document.createElement('iframe');
     iframe.style.display = 'none';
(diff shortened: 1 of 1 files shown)


guybedford:
Looks great, thank you for the fix. Published in 0.10.7.

PR #142, opened 2021-06-30
Use `link[rel=modulepreload]` shim only when necessary

tomayac:
Adds feature detection to figure out browser support.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -13,6 +13,7 @@ export function createBlob (source, type = 'text/javascript') {
 
 export const hasDocument = typeof document !== 'undefined';
 
+export const supportsModulePreload = document.createElement('link').relList.supports('modulepreload');
 // support browsers without dynamic import support (eg Firefox 6x)
 export let supportsDynamicImport = false;
 export let supportsJsonAssertions = false;
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -7,6 +7,7 @@ import {
   resolveIfNotPlainOrUrl,
   resolvedPromise,
   dynamicImport,
+  supportsModulePreload,
   supportsDynamicImport,
   supportsImportMeta,
   supportsImportMaps,
@@ -315,8 +316,10 @@ async function processScripts () {
     clearTimeout(waitingForImportMapsInterval);
     waitingForImportMapsInterval = 0;
   }
-  for (const link of document.querySelectorAll('link[rel="modulepreload"]'))
-    processPreload(link);
+  if (!supportsModulePreload) {
+    for (const link of document.querySelectorAll('link[rel="modulepreload"]'))
+      processPreload(link);
+  }
   for (const script of document.querySelectorAll('script[type="module-shim"],script[type="importmap-shim"],script[type="module"],script[type="importmap"]'))
     await processScript(script);
 }
@@ -327,7 +330,7 @@ new MutationObserver(mutations => {
     for (const node of mutation.addedNodes) {
       if (node.tagName === 'SCRIPT' && node.type)
         processScript(node, !firstTopLevelProcess);
-      else if (node.tagName === 'LINK' && node.rel === 'modulepreload')
+      else if (!supportsModulePreload && node.tagName === 'LINK' && node.rel === 'modulepreload')
         processPreload(node);
     }
   }


guybedford:
Thanks for the PR. One thing I would just like to test before landing this is if the modulepreload tag is able to prime the fetch cache for the loader when running in shim mode. Otherwise this will entirely disable preloading for shim mode in modern Chrome.

Will leave this open until I've had time to test that, or if you are able to investigate further please do share results.

tomayac:
Looking at your code, you check for `link[rel=modulepreload]`, but don't monkey-patch `HTMLLinkElement.relList` to report it supports `'modulepreload'`, which is what my code checks. So in shim mode the variable `supportsModulePreload` is `false`, which means your shim code paths would be taken.

guybedford:
I appreciate that, but my concern is that even in browsers that have full support for modulepreload, they might not support import maps.

So the specific feature I want to ensure that we keep is that running in shim mode to polyfill import maps, where es-module-shims uses `fetch` to load all the modules, that the fetch cache will still be primed by `modulepreload` properly. A very simple test should be able to confirm this.

tomayac:
I just ran the following code in Safari:

```html
<link rel="modulepreload" href="./nested/app.js">

<script async src="es-module-shims.js"></script>

<script type="importmap-shim">
  {
    "imports": {
      "app": "./nested/app.js"
    }
  }
</script>

<script type="module-shim">import 'app'</script>
```

Screen Shot 2021-08-06 at 16 14 14

The way I read this is that the module preload is working and the second request is served by the disk cache.

guybedford:
Can you verify the same using dynamic import, and in Chrome with cache disabled?

guybedford:
(and with a delay on the dynamic import)

guybedford:
The `app.js` import in the second example doesn' look like it is coming from the cache to me, so that was my concern with this PR yes.

tomayac:
You asked me to disable the cache. With cache enabled, it looks perfect:

Screen Shot 2021-08-06 at 16 31 12

guybedford:
Disk cache seems odd to me though, we should verify it's not cached from a previous run. Can you share your test case, then I can take a further look.

tomayac:
```html
<link rel="modulepreload" href="./nested/app.js">

<script async src="es-module-shims.js"></script>

<script type="importmap-shim">
  {
    "imports": {
      "app": "./nested/app.js"
    }
  }
</script>

<script type="module-shim">import 'app'</script>
```

guybedford:
Sure, let me take 5 minutes out of my day to do this for you.

tomayac:
With an empty cache, this is what it looks like:

Screen Shot 2021-08-06 at 16 36 40

Not sure why, though…

guybedford:
Here's my test on master with caching disabled

image

Note that there isn't a second request.

So it took me 5 minutes to verify my original assumption - this was your job to do.

guybedford:
Closing for reasons described above. Happy to reconsider if there is sufficient evidence for a change.

tomayac:
>>> Will leave this open until 👉I👈've had time to test that, or if you are able to investigate further please do share results.

>>  Can you share your test case, then 👉I👈 can take a further look.

> So it took me 5 minutes to verify my original assumption.

Sorry to have wasted your precious time.

guybedford:
It's best to focus on the direct issue at hand than to get caught up in personal issues. This is open source it's about the work at the end of the day.

As I've said, if there's evidence there are benefits to this change I'm open to it. Apologies for my tone, but I hope you can agree with my technical analysis though - and if you disagree I'm happy to discuss further, but it's unfair to place all the onus on maintainers - a certain amount of effort in response is required as the cost of getting your contribution through.

tomayac:
Tone apologies accepted, and no offense taken. And honest apologies from my end, I did not fully grasp what shim mode does before opening this PR.

guybedford:
All good! Ok just looking through the scenario again - 

> but don't monkey-patch HTMLLinkElement.relList to report it supports 'modulepreload', which is what my code checks

Should we be doing this monkey patching?

tomayac:
> Should we be doing this monkey patching?

I think it would make sense, since you're de facto adding support for the feature.

guybedford:
Ok, posted https://github.com/guybedford/es-module-shims/issues/147.

PR #144, opened 2021-07-18
Support processing nested script / link tags in MutationObserver

dbackeus:
Previously `<script type="module-shim">` tags appended to DOM when replacing the entire `<body>` tag using eg. [Turbolinks](https://github.com/turbolinks/turbolinks) would not be processed.

I have confirmed this patch resolves the issues I had with `es-module-shims` in a local Turbolinks enabled Ruby on Rails application.

I briefly looked at the test folder but failed to figure out how I could add tests for this behaviour. Seems like there are no tests for the MutationObserver currently (no tests failed when I attempted to remove the MutationObserver).

Let me know if there is a strong desire to test this diff. If so I'll need some help to get started.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -329,6 +329,14 @@ new MutationObserver(mutations => {
         processScript(node, !firstTopLevelProcess);
       else if (node.tagName === 'LINK' && node.rel === 'modulepreload')
         processPreload(node);
+      else if (node.querySelectorAll) {
+        for (const script of node.querySelectorAll('script[type="module-shim"],script[type="importmap-shim"],script[type="module"],script[type="importmap"]')) {
+          processScript(script, !firstTopLevelProcess);
+        }
+        for (const link of node.querySelectorAll('link[rel=modulepreload]')) {
+          processPreload(link);
+        }
+      }
     }
   }
 }).observe(document, { childList: true, subtree: true });


guybedford:
Thanks for posting the PR. I'm trying to understand the exact scenario in play here. Can you share a replication or end to end example of how this case can happen?

dbackeus:
Since it's 2021 and I seem to have left my local friendly FTP in the 90s somewhere the best I can do right this moment is to hand you this html code, but you'll have to run it locally yourself, eg. via `python -m SimpleHTTPServer` if you have python installed...

Add to `turbolinks-es-module-shims-1.html`:
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Page 1</title>

    <script defer src="https://cdn.jsdelivr.net/npm/turbolinks@5.2.0/dist/turbolinks.js"></script>
    <meta name="turbolinks-cache-control" content="no-cache">

    <script async src="https://unpkg.com/es-module-shims@0.12.1/dist/es-module-shims.js"></script>
  </head>
  <body>
    <h1>Page 1</h1>
    <script type="module-shim">
     console.log("page 1: Hello from module-shim!")
   </script>
   <script type="module">
     console.log("page 1: Hello from module!")
   </script>
   <a href="turbolinks-es-module-shims-2.html">Page 2</a>
  </body>
</html>

```

Add to `turbolinks-es-module-shims-2.html`:
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Page 2</title>

    <script defer src="https://cdn.jsdelivr.net/npm/turbolinks@5.2.0/dist/turbolinks.js"></script>
    <meta name="turbolinks-cache-control" content="no-cache">

    <script async src="https://unpkg.com/es-module-shims@0.12.1/dist/es-module-shims.js"></script>
  </head>
  <body>
    <h1>Page 2</h1>
    <script type="module-shim">
     console.log("page 2: Hello from module-shim!")
   </script>
   <script type="module">
     console.log("page 2: Hello from module!")
   </script>
   <a href="turbolinks-es-module-shims-1.html">Page 1</a>
  </body>
</html>
```

Now run eg. `python -m SimpleHTTPServer` to serve the directory you put the files in and open http://localhost:8000/turbolinks-es-module-shims-1.html in Chrome. You'll notice that on first load you'll get console output from both module-shim and module. However when you click the link to navigate to Page 2 the module-shim will not execute because the es-modules-shim MutationObserver doesn't notice the insertion of the script tag which is nested inside of the new body tag.

If you're unfamiliar with Turbolinks it essentially hijacks `<a href>` tags and changes their behaviour to fetch the destination page using ajax and then mutates the DOM with the results to improve performance and bring the user experience closer to a "single page app". It's enabled by default on all Ruby on Rails applications.

guybedford:
Thanks I was able to verify the scenario and that makes complete sense. Will post this out in the next few days.

src/es-module-shims.js
```diff
@@ -329,6 +329,14 @@ new MutationObserver(mutations => {
         processScript(node, !firstTopLevelProcess);
       else if (node.tagName === 'LINK' && node.rel === 'modulepreload')
         processPreload(node);
+      else if (node.querySelectorAll) {
+        for (const script of node.querySelectorAll('script[type="module-shim"],script[type="importmap-shim"],script[type="module"],script[type="importmap"]')) {
```
guybedford, line 333: We could possibly simplify this like the above to just:

```suggestion
        for (const script of node.querySelectorAll('script')) {
```

and then let the function handle the rest.

dbackeus: I used the same selector as in https://github.com/guybedford/es-module-shims/blob/main/src/es-module-shims.js#L320 Another option could be to store the selector in a `const` and reuse it.

Feel free to adapt the diff how you prefer it. I don't have a strong opinion either way :)

dbackeus:
Thanks for the merge!

Can we get a patch release on NPM for this as well?

guybedford:
Np, published in 0.12.2.

guybedford:
@dbackeus I'm planning to revert this in https://github.com/guybedford/es-module-shims/pull/189, see my description of the change there. Happy to discuss further.

Issue #143, opened 2021-07-09
Resolver hook


guybedford:
It could be useful to expose a resolver API via:

```js
globalThis.esmsInitOptions = {
  resolve: (specifier, parent, defaultResolve) => defaultResolve(specifier, parent)
}
```

where `defaultResolve` is the normal resolve implementation of URL and import map resolution.

This would make it easy to eg hook stripping querystrings:

```js
globalThis.esmsInitOptions = {
  resolve: (specifier, parent, defaultResolve) => defaultResolve(specifier.split('?')[0], parent)
}
```

zhoukekestar:
Any update on this feature ? It seems simple to implement and very useful.

guybedford:
@zhoukekestar if you're interested in working on this a PR would be great.

guybedford:
Released and implemented in https://github.com/guybedford/es-module-shims#resolve-hook by @zhoukekestar.

PR #146, opened 2021-08-02
Support resolve hook 

zhoukekestar:
Support resolve hook. For #143

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -393,13 +393,16 @@ function processPreload (link) {
   fetch(link.href, fetchOpts);
 }
 
+function defaultResolve(id, parentUrl) {
+  return resolveImportMap(importMap, resolveIfNotPlainOrUrl(id, parentUrl) || id, parentUrl);
+}
 
 async function resolve(id, parentUrl) {
   let urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);;
 
   let resolved;
   if (esmsInitOptions.resolve) {
-    resolved = await esmsInitOptions.resolve(id, parentUrl, resolveIfNotPlainOrUrl);
+    resolved = await esmsInitOptions.resolve(id, parentUrl, defaultResolve);
   }
   else {
     resolved = resolveImportMap(importMap, urlResolved || id, parentUrl);
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -281,7 +281,7 @@ function getOrCreateLoad (url, fetchOpts, source) {
 
   load.L = load.f.then(async () => {
     let childFetchOpts = fetchOpts;
-    load.d = await Promise.all(load.a[0].map(async ({ n, d, a }) => {
+    load.d = (await Promise.all(load.a[0].map(async ({ n, d, a }) => {
       if (d >= 0 && !supportsDynamicImport ||
           d === 2 && (!supportsImportMeta || source.slice(end, end + 8) === '.resolve') ||
           a && !supportsJsonAssertions)
@@ -297,8 +297,7 @@ function getOrCreateLoad (url, fetchOpts, source) {
       if (childFetchOpts.integrity)
         childFetchOpts = Object.assign({}, childFetchOpts, { integrity: undefined });
       return getOrCreateLoad(r, childFetchOpts).f;
-    }));
-    load.d = load.d.filter(l => l);
+    }))).filter(l => l);
   });
 
   return load;
@@ -396,17 +395,16 @@ function processPreload (link) {
 
 
 async function resolve(id, parentUrl) {
-  let urlResolved;
+  let urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);;
 
+  let resolved;
   if (esmsInitOptions.resolve) {
-    urlResolved = await esmsInitOptions.resolve(id, parentUrl, resolveIfNotPlainOrUrl);
+    resolved = await esmsInitOptions.resolve(id, parentUrl, resolveIfNotPlainOrUrl);
+  }
+  else {
+    resolved = resolveImportMap(importMap, urlResolved || id, parentUrl);
   }
 
-  if (!urlResolved) {
-    urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);
-  }
-
-  const resolved = resolveImportMap(importMap, urlResolved || id, parentUrl);
   return { r: resolved, m: urlResolved !== resolved };
 }
 
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -91,7 +91,8 @@ async function importShim (id, parentUrl = pageBaseUrl, _assertion) {
   await featureDetectionPromise;
   // Make sure all the "in-flight" import maps are loaded and applied.
   await importMapPromise;
-  return topLevelLoad(resolve(id, parentUrl).r || throwUnresolved(id, parentUrl), { credentials: 'same-origin' });
+  const resolved = await resolveHook(id, parentUrl);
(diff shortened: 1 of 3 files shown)


guybedford:
Thanks for working on this. Is there a specific reason you're looking to implement an async hook or would a sync one also work? That could also simplify the implementation in folding it into the current resolver function.

zhoukekestar:
My first version is just a sync function too. But if we want some dynamic components, maybe we need to fetch its dependencies asynchronously & lazily. 

For example: we want to render a WebComponent but have no dependencies (and import map) when the page loaded. Trigger ( iframe page or something ) and render this WebComponent (Dialog) lazily when we need it may be our expectation.

Maybe it's not the best practice. But someone may need this feature in the future. 😀

guybedford:
Yeah I'm open to either. But either way I would still rather fold the resolver hook into a single call in a single resolve function.

zhoukekestar:
Then how about the API parameter `defaultResolver` in `esmsInitOptions.resolve(id, parentUrl, defaultResolver);`?
or Just scope the defaultResolver inside the resolveHook like:

```js
async function resolveHook(id, parentUrl) {
  function defaultResolver(id, parentUrl) {}

  if ...
}
```

src/es-module-shims.js
```diff
@@ -386,12 +389,31 @@ function processPreload (link) {
...
+    }
+
+    return resolved;
+  }
+
+  return defaultResolver(id, parentUrl);
+}
+
+function defaultResolver (id, parentUrl) {
   const urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);
```
guybedford, line 417: This `urlResolved` should be moved into the top of the `resolveHook` as it is supposed to be the comparison of the browser resolver, to determine if import maps are used. Then `defaultResolver` as referenced by the resolve hook will return a string nicely too.

src/es-module-shims.js
```diff
@@ -386,12 +389,31 @@ function processPreload (link) {
...
   const fetchOpts = getFetchOpts(link);
-  // save preloaded fetch options for later load  
+  // save preloaded fetch options for later load
   fetchOptsMap.set(link.href, fetchOpts);
   fetch(link.href, fetchOpts);
 }
 
-function resolve (id, parentUrl) {
+
+async function resolveHook(id, parentUrl) {
```
guybedford, line 398: Can we just call this `resolve`. I actually misread initially, I think just the rename clarifies my original confusion.

src/es-module-shims.js
```diff
@@ -386,12 +389,31 @@ function processPreload (link) {
...
+    return resolved;
+  }
+
+  return defaultResolver(id, parentUrl);
+}
+
+function defaultResolver (id, parentUrl) {
   const urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);
   const resolved = resolveImportMap(importMap, urlResolved || id, parentUrl);
   return { r: resolved, m: urlResolved !== resolved };
```
guybedford, line 419: This part can then also be moved up into the main resolve function.

guybedford (approved):

zhoukekestar:
Just updated the PR, simplify the resolve hook.

src/es-module-shims.js
```diff
@@ -295,7 +297,8 @@ function getOrCreateLoad (url, fetchOpts, source) {
       if (childFetchOpts.integrity)
         childFetchOpts = Object.assign({}, childFetchOpts, { integrity: undefined });
       return getOrCreateLoad(r, childFetchOpts).f;
-    }).filter(l => l));
+    }));
+    load.d = load.d.filter(l => l);
```
guybedford, line 301: Rather than reassigning, we can just run the filter on a bracket expression around the await.

src/es-module-shims.js
```diff
@@ -386,13 +389,23 @@ function processPreload (link) {
...
+  if (esmsInitOptions.resolve) {
+    urlResolved = await esmsInitOptions.resolve(id, parentUrl, resolveIfNotPlainOrUrl);
+  }
+
+  if (!urlResolved) {
+    urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);
+  }
+
   const resolved = resolveImportMap(importMap, urlResolved || id, parentUrl);
   return { r: resolved, m: urlResolved !== resolved };
```
guybedford, line 410: I know it's not necessarily clear what the purpose of urlResolved is, but the following should better maintain the current intent:

```suggestion
  const urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);

  let resolved;
  if (esmsInitOptions.resolve) {
    resolved = await esmsInitOptions.resolve(id, parentUrl, resolveIfNotPlainOrUrl);
  }
  else {
    resolved = resolveImportMap(importMap, resolved || id, parentUrl);
  }

  return { r: resolved, m: urlResolved !== resolved };
```

guybedford: Finally I think we should pass the full original resolver as the third argument to `esmsInitOptions.resolve` not just the `resolveIfNotPlainOrUrl` - that way the resolver hook has access to import map resolutions.

guybedford (commented):
Thanks for your work on this. Still not quite there though, see the feedback further.

src/es-module-shims.js
zhoukekestar: Pass all context to the third argument?
```js
esmsInitOptions.resolve(id, parentUrl, { resolveIfNotPlainOrUrl, resolveImportMap, importMap });
```
So the resolver hook can reuse this?

guybedford:
The approach looks good, thanks for the work on this. See the last point about the nature of the `defaultResolve` function passed.

src/es-module-shims.js
```diff
@@ -386,14 +388,26 @@ function processPreload (link) {
...
 
-function resolve (id, parentUrl) {
-  const urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);
-  const resolved = resolveImportMap(importMap, urlResolved || id, parentUrl);
+function defaultResolve(id, parentUrl) {
+  return resolveImportMap(importMap, resolveIfNotPlainOrUrl(id, parentUrl) || id, parentUrl);
+}
+
+async function resolve(id, parentUrl) {
+  let urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);;
```
guybedford, line 401:
```suggestion
  let urlResolved = resolveIfNotPlainOrUrl(id, parentUrl);
```

guybedford (approved):
Looks great, think we've cracked it.

Re the sync / async thing this also relates to what browsers will expose via `import.meta.resolve`. See eg https://github.com/whatwg/html/issues/3871 for discussion. Chrome currently wants to implement a sync resolver I believe. If we permit an async resolver we may be constrained in future if `import.meta.resolve` turns out to go sync. But we can just make a breaking change in this project too I suppose.

test/test-resolve-hook.html
```diff
@@ -0,0 +1,48 @@
...
+  assert.ok = assert;
+  function equal (a, b) {
+    if (a !== b)
+      throw new Error('Expected "' + a + '" to be "' + b + '"');
+  }
+  self.fail = function (msg) {
+    throw new Error(msg);
+  };
+
+  const suites = ['resolve-hook'];
```
guybedford, line 34: Rather than having a separate test-resolve-hook, can we just add this to the main suite in test/test.html? That will ensure it's captured by CI.

guybedford:
Landed in https://github.com/guybedford/es-module-shims/commit/1301af1ed7ceefea175cb6e4a7d4214696a9cde2, test refactoring and Firefox 60 fixes in https://github.com/guybedford/es-module-shims/commit/c6ee67226864569b3d8b325957db3e41a41ccc26.

zhoukekestar:
Thanks, 🎉

PR #166, opened 2021-08-27
Avoids `eval()` to allow stricter CSP

Lcfvs:
Logically, we don't need an `eval()`, since if `import()` isn't callable as a function, the call to `dynamicImport()` throws **synchronously**

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -19,7 +19,8 @@ export let supportsJsonAssertions = false;
 export let supportsCssAssertions = false;
 export let dynamicImport;
 try {
-  dynamicImport = (0, eval)('u=>import(u)');
+  dynamicImport = u=>import(u);
+  dynamicImport(createBlob(''))
   supportsDynamicImport = true;
 }
 catch (e) {
@@ -183,7 +184,7 @@ function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap) {
   for (let p in packages) {
     const resolvedLhs = resolveIfNotPlainOrUrl(p, baseUrl) || p;
     let target = packages[p];
-    if (typeof target !== 'string') 
+    if (typeof target !== 'string')
       continue;
     const mapped = resolveImportMap(parentMap, resolveIfNotPlainOrUrl(target, baseUrl) || target, baseUrl);
     if (mapped) {


guybedford:
Perhaps try something like this:

```js
{
  let err;
  self.addEventListener('error', e => err = e.error);
  const syntheticDynamicImport = blobUrl => {
    const topLevelBlobUrl = createBlob(
      `import*as m from'${blobUrl}';self._esmsi=m;`
    );
    const s = document.createElement('script');
    s.type = 'module';
    s.src = topLevelBlobUrl;
    document.head.appendChild(s);
    return new Promise((resolve, reject) => {
      s.addEventListener('load', () => {
        document.head.removeChild(s);
        if (self._esmsi) {
          resolve(self._esmsi, baseUrl);
          self._esmsi = null;
        }
        else {
          reject(err);
        }
      });
    });
  };
  dynamicImport = syntheticDynamicImport('export let d = u => import(u)')
  .then(({ d }) => {
    // dynamic import exists, use it
    return dynamicImport = d;
  }, () => {
    // does not exist -> use synthetic
    return dynamicImport = syntheticDynamicImport;
  });
}
```

Lcfvs:
Hi,

I just compared [the static `import` support](https://caniuse.com/#feat=es6-module) VS [the dynamic `import()` support](https://caniuse.com/es6-module-dynamic-import)

It seems to imply to have the same users... then, couldn't we legitimately drop the Firefox 60 support? (it can avoid the need to add some script tags, to listen the global errors, ...)

guybedford:
I don't think it's worth justifying dropping 0.4% of browsers just to avoid ~30 lines of code.

Lcfvs:
Not really for the code length, but the DOM mutations & the global errors listening.

But it's ok, I'm just wondering ;)

Lcfvs:
Still on the proposals, I thinking about an idea.

What do you think about to have 2 shims?

A shim close to the current version, with the same support and another one a few stricter & lightest than the previous one, but since Firefox 64.

Of course, it implies a split of the current code in small modules to reuse the common things into the 2 versions.

guybedford:
> What do you think about to have 2 shims?

We've already discussed this - ~30 lines of code isn't worth putting this overhead of consideration onto users.

Lcfvs:
Really, it isn't about the size but because the error handling is really different.

To detail what I mean...

Using a `syntheticDynamicImport()`, we're forced to listen globally about any error, then when we `reject(error)` [here](https://github.com/guybedford/es-module-shims/blob/main/src/common.js#L45), the error isn't reliable, due to the async things we can have anywhere or due to "externally" loaded scripts. We can't be sure about the origin of the error, then it can have some inconsistent/unstable behaviors. 

Using a true dynamic `import()`, if there is an error, we have the guarantee about the error origin (the module itself or a dependency).

I understand your wish to cover the largest users panel but, it's a few like the famous NPM warning `I'll try to do my best with it!`, then yes, it can work as expected but it can also have an unpredictable behavior.

guybedford:
> We can't be sure about the origin of the error, then it can have some inconsistent/unstable behaviors.

No, this is incorrect. There are consistent sync state guarantees around the load event of scripts, which is how SystemJS is able to work at all as a project using the same approach.

Lcfvs:
Ok, ok, I didn't know that, sorry then :)

guybedford:
Thanks for poking on this! This has now been superseded by https://github.com/guybedford/es-module-shims/pull/183.

PR #170, opened 2021-09-06
feat: script events

guybedford:
This adds script events for individual module scripts in the page.

In polyfill mode, load events are triggered for both success and failure per the spec, resulting in two load events always applying - one from the native loader and one from the polyfill.

In shim mode, there will always only be one load event for success or failure.


guybedford:
After further testing it turns out:

* The "load" event is **always fired** for errors and non errors in all browsers per the spec.
* The "error" event is **never thrown** for module scripts at all per the spec and in all browsers.
* Thus the only way to detect errors is to listen to the window.onerror method.

I've updated the PR and description based on this.

guybedford:
And on further thought, I think in polyfill mode it makes sense to just always retrigger the load event so there will always be two load events. This is much simpler for a user to check than trying to determine whether or not it is a genuine retrigger.

PR #172, opened 2021-09-06
Polyfill enabling

guybedford:
I wasn't quite happy with the `skip-shim` concept in https://github.com/guybedford/es-module-shims/pull/171 - it creates too much unnecessary overhead for users.

Instead the approach taken in this PR is to disable CSS modules and JSON modules automatic polyfilling, and instead require those to be explicitly opted-in to. To avoid compatibility issues we then throw whenever these are loaded in unsupported browsers to avoid the spotty compat issues that would make compat testing harder.

This way we maintain the baseline native import maps target without degradation and users who want newer features can explicitly indicate that. 

Next year, we can switch these on for the baseline by default, and in this way gradually adjust the target via the release schedule of the project itself to ensure optimal native performance for users.

The option is:

```html
<script>
window.esmsInitOptions = { polyfillEnable: ['css-modules', 'json-modules'] };
</script>
```

to use these newer features in polyfill mode, which I think is much clearer for users from an intuition and intent perspective as well.


dmnsgn:
@guybedford would it be possible/make sense to add "dynamic-import-maps" as a "polyfillEnable" option as well since it is in the spec even though implemented in browsers?
It is very often the only feature I need and it would allow the following code to be valid:

```html
<script>
window.esmsInitOptions = { polyfillEnable: ['dynamic-import-maps'] }
</script>
<script src="web_modules/import-map.json" type="importmap"></script>
<script async src="web_modules/es-module-shims.js" type="module"></script>

<script type="module">
import myModule from "bare-specifier-from-autogenerated-import-map";
</script>
```

guybedford:
@dmnsgn `polyfillEnable` only includes features that are natively supported. Since dynamic import maps is not even specified it isn't a feature that can fall under the polyfill unfortunately, so is only available in shim mode. The other reason for this is to ensure users don't mistakenly think they're just using ES Module Shims as a polyfill even when it's never hitting a native path. As soon as there is native intent and movement on the feature we will automatically include it in polyfill mode though. Rather bump the native issues if you want this faster :)

dmnsgn:
Thanks, makes sense! I hoped there was more interest for it but I have struggled finding the native issues (no mention on [chromestatus.com](https://chromestatus.com/features) or https://webkit.org/status) so not sure where to push 😅

guybedford:
The spec issue is here -
https://github.com/WICG/import-maps/issues/235

PR #173, opened 2021-09-07
feat: align map constraints with native

guybedford:
This aligns _polyfill mode_ to restrict the features it provides for import maps, in order to ensure that the native subset is treated as the primary target, and users don't inadvertently use behaviours that cause the polyfill to run unnecessarily on modern browsers.

These restrictions are:

1. Support for multiple import maps on the page is removed.
2. Support for import maps with a `"src"` attribute is removed.
3. Support for dynamically injecting an import map after any modules has loaded on the page is removed.

In addition we fully restrict the `resolve` and `fetch` hooks to only apply in shim mode, and be ignored otherwise.

The import map polyfill check is also simplified to only check for bare specifier maps (which fail) and not arbitrary URL maps as indicating re-execution under the polyfill.

All of the above features remain fully supported in shim mode - therefore applications that want to take advantage of these behaviours that have zero native support need to fully opt-in to shim mode, therefore fully understanding that they are not getting any native executions.

In future we could reenable these features when supported natively (user pressure where are you!!!???), via the new `polyfillEnable` system which could shift to the baseline features over time.


guybedford:
As a big change, this will be the main aspect of the 0.13.0 release.

PR #189, opened 2021-09-18
fix: revert nested mutation observers

guybedford:
This reverts https://github.com/guybedford/es-module-shims/pull/144. After investigating turbolinks more thoroughly, it turns out script injections are separately applied to the head in https://github.com/hotwired/turbo/blob/c4891515645880ae1f9b4ee5757a8d730b89ac4e/src/core/drive/page_renderer.ts#L60 and in the body in https://github.com/hotwired/turbo/blob/c4891515645880ae1f9b4ee5757a8d730b89ac4e/src/core/frames/frame_renderer.ts#L48.

This is consistent with the fact that in HTML, setting the `body` `innerHTML` does not cause script execution.

#144 causes a huge amount of performance cost due to querying all mutations so this is a significant performance improvement.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -479,14 +479,6 @@ new MutationObserver(mutations => {
         processScript(node, !firstTopLevelProcess);
       else if (node.tagName === 'LINK' && node.rel === 'modulepreload')
         processPreload(node);
-      else if (node.querySelectorAll) {
-        for (const script of node.querySelectorAll('script[type="module-shim"],script[type="importmap-shim"],script[type="module"],script[type="importmap"]')) {
-          processScript(script, !firstTopLevelProcess);
-        }
-        for (const link of node.querySelectorAll('link[rel=modulepreload]')) {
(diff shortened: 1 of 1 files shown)


dbackeus:
I haven't played around with the new `turbo` library that you are referring to here, which is kind of the successor to `turbolinks` which behaviour I used when testing #144. So perhaps `turbo` doesn't have the same issue but can't confirm.

The point of #144 was to remove the discrepancy between how Chrome natively behaved with `type="module"` vs how `es-module-shims` behaved via `type="module-shim"` in this scenario. Seemed to me that there was inconsistency between native and shim, so not sure if we're trading off consistency for performance here.

guybedford:
@dbackeus the issue is that `document.body.innerHTML = "<script>console.log('exec')</script>"` does not result in native script execution, so while it can work due to the way es module shims picks up scripts, it is not recommended as the execution mechanism. That is consistency should be around following patterns that work natively then making them work consistently for the polyfill.

Issue #211, opened 2021-10-10
Checking existing import map before injecting new import map


lewisl9029:
Hi there! 

I've been trying to iron out some of the edge cases in my react-refresh workflow using es-module-shims's fetch hook. One of these requires injecting new import maps for new files that were added during development. 

To avoid errors (according to the spec, dynamically injected import maps must not override any existing entry) and wasted work, I'd like to be able to only inject new import maps if the files added don't already exist. I could probably hack together something myself by reading the importmap/importmap-shim scripts on the page and merging them, but since es-module-shims already keeps track of the latest import map state, it would be perfect if I could just read from that.

As far as I can tell, es-module-shims doesn't currently expose any global runtime APIs outside of the `importShim` function, so I wanted to get your thoughts/preferences on what this API should look like before I go ahead and try to open a PR. The actual implementation could be as trivial as replacing https://github.com/guybedford/es-module-shims/blob/main/src/es-module-shims.js#L66 with `self.importMap = // ...` so it's accessible globally, but I imagine we'd want some namespacing and/or maybe expose a function instead to better control access to the internal state.

Or maybe you'd prefer to avoid exposing internal import map state altogether? In that case, maybe we could refactor out a few functions for users to call to reuse the same logic that es-module-shim uses to build up our own import map state? Barring that, I'd probably just have to resort to copy pasting that logic from es-module-shims, which tbh isn't the end of the world either.

Let me know which path you'd prefer!

lewisl9029:
Yep that's exactly it!

guybedford:
As much as possible, the starting point for this project is the spec side of things. So implementing behaviours that are non-spec should only ever apply to shim mode for the most part. I'm very hesitant to add anything new to polyfill mode.

In terms of the spec behaviour there are definitely uncertainties:

* Would multiple import maps throw or silently override?
* Perhaps silent overrides can be permitted for import mappings that have not yet been "used"

I tend to think throwing overall is a safer approach to any spec work here though.

Given that, I would try to come at it from the perspective of how you would write your app if this was native HTML. (3) seems to be the only option that follows that. And then having the correct error behaviours likely isn't a blocker to your work anyway.

PR #213, opened 2021-10-12
Read current import map

lewisl9029:
Implements the changes discussed in #211.

- We now throw when detecting a potential override of existing import map entries. The current implementation covers the fairly trivial case for `imports` overrides as described in the [extension proposal](https://github.com/guybedford/import-maps-extensions#1-defining-immutable-import-map-extension), but not the `scopes` case, which seems like it could get a lot more tricky. Not super confident I could adequately address all the potential edge cases there, so I've decided to leave that part for @guybedford to potentially tackle as a follow up for another day. :)

  - Given this error behavior could break code that depended on being able to dynamically override import map entries successfully, we may want to consider bumping the major version to drive attention to the potential breaking change.

- We also expose 2 new global functions `getCurrentImportMap`, `getNextImportMap`, that allow users to reuse the same logic es-module-shims uses internally to compute import map state from scratch, without exposing any internal state from es-module-shims directly or indirectly, as we discussed in #211. Longer term, I suspect we may even want to pull this logic out into its own independent npm module and have es-module-shims just import that module as any other user would.

- I've also added a new subsection in the readme to document the new error behavior on dynamic import map overrides, along with some small examples of how to use the new functions to avoid running into these errors.

Happy to make any changes needed to get this ready to ship. Cheers!

### DIFF
--- a/test/shim.js
+++ b/test/shim.js
@@ -383,7 +383,7 @@ suite('Errors', function () {
 
     removeImportMap();
 
-    assert(error.message === 'Dynamic import map rejected: Attempted to override existing import map entry at `global1` from value `http://localhost:8080/test/fixtures/es-modules/global1.js` to `data:text/javascript,throw new Error(\'Shim should not allow dynamic import map to override existing entries\');`.');
+    assert(error.message === 'Dynamic import map rejected: Overrides entry "global1" from http://localhost:8080/test/fixtures/es-modules/global1.js to data:text/javascript,throw new Error(\'Shim should not allow dynamic import map to override existing entries\');.');
   })
 
   function insertDynamicImportMap(importMap) {


src/common.js
```diff
@@ -126,6 +126,9 @@ export function resolveUrl (relUrl, parentUrl) {
 function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap) {
   for (let p in packages) {
     const resolvedLhs = resolveIfNotPlainOrUrl(p, baseUrl) || p;
+    if (outPackages[resolvedLhs]) {
+      throw new Error(`Attempted to override existing import map entry at \`${resolvedLhs}\` from value \`${outPackages[resolvedLhs]}\` to \`${packages[resolvedLhs]}\`.`);
```
guybedford, line 130: Perhaps let's clarify we are ignoring the entire import map:

> Dynamic import map rejected: Attempted to override existing import map entry "${resolvedLhs}" from "${...}" to "${...}".

package.json
```diff
@@ -12,15 +12,16 @@
...
     "test:test-base-href": "cross-env TEST_NAME=test-base-href node test/server.mjs",
     "test:test-csp": "cross-env TEST_NAME=test-csp node test/server.mjs",
     "test:test-early-module-load": "cross-env TEST_NAME=test-early-module-load node test/server.mjs",
     "test:test-polyfill": "cross-env TEST_NAME=test-polyfill node test/server.mjs",
     "test:test-polyfill-wasm": "cross-env TEST_NAME=test-polyfill-wasm node test/server.mjs",
     "test:test-preload-case": "cross-env TEST_NAME=test-preload-case node test/server.mjs",
     "test:test-revoke-blob-urls": "cross-env TEST_NAME=test-revoke-blob-urls node test/server.mjs",
-    "test:test-shim": "cross-env TEST_NAME=test-shim node test/server.mjs"
+    "test:test-shim": "cross-env TEST_NAME=test-shim node test/server.mjs",
+    "test:test-get-import-maps": "cross-env TEST_NAME=test-get-import-maps node test/server.mjs"
```
guybedford, line 24: Rather than a separate test, let's just fold this entirely into test-shim. These are more the "suites" than the tests...

src/es-module-shims.js
```diff
@@ -506,3 +518,12 @@ function processPreload (link) {
...
 }
+
+
+self.getNextImportMap = getNextImportMap
+
+self.getCurrentImportMap = () => {
+  return Array.from(getImportMapScripts()).reduce((gettingPreviousImportMaps, script) => {
+    return gettingPreviousImportMaps.then((importMap) => getNextImportMap(script, importMap))
+  }, Promise.resolve(emptyImportMap))
+}
```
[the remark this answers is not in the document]
lewisl9029: Implemented here: https://github.com/guybedford/es-module-shims/pull/213/commits/0edb5d0e16ef36e3243d574d6981d38995bf1352

Issue #217, opened 2021-11-03
`sourceMappingURL` and `sourceURL` rewriting breaks my apps


hronro:
https://github.com/guybedford/es-module-shims/blob/1.3.0/src/es-module-shims.js#L262-L264

I don't think a simple regex replacement is enough, please consider the code below:

```js
// this function usually comes from a bundler runtime
function foo(modulePath){
  bar("// # sourceURL="+modulePath);
}
```

Currently es-module-shims will rewrite this function into invalid syntax.

Perhaps we should parse the whole javascript and only replace `sourceMappingURL` and `sourceURL` in comments? (maybe I'm wrong, but it seems it have already parsed javascript in `es-module-lexer`?)

guybedford:
The sourcemapping regex is supposed to stop on newlines and only detect source map comments at the end of the source.

I've added a possible fix to the regular expression in https://github.com/guybedford/es-module-shims/pull/218 - can you perhaps let me know if that resolves the issue for you?

PR #218, opened 2021-11-03
Fix sourcemapping regex

guybedford:
Resolves https://github.com/guybedford/es-module-shims/issues/217.

//cc @hronro

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -259,9 +259,9 @@ function resolveDeps (load, seen) {
     resolvedSource += source.slice(lastIndex);
   }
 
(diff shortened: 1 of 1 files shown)


hronro:
CleanShot 2021-11-04 at 11 31 05@2x

It's really hard to be done properly with simple regex replacement.

guybedford:
Because they have to be at the end of the source it is actually quite well defined.

I've pushed up a new fix, let me know if that can work for you?

hronro:
Cook, the new fix works for me.

guybedford:
Thanks for confirming, I will include in the next release in the next few days.

PR #223, opened 2021-11-16
Fixed dynamic retrieval of nonce attribute 

MathiasWP:
I got this error:
Screenshot 2021-11-16 at 11 22 24

Which was thrown here:
Screenshot 2021-11-16 at 11 23 52

So i tried to check how the nonce value was retrieved:
https://github.com/guybedford/es-module-shims/blob/b99eddfe33445cff16896f4353d66e3c7a001a0f/src/options.js#L17

Which violates the rules for accessing nonce: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce#accessing_nonces_and_nonce_hiding

Changing the line from 
```ts
nonce = nonceElement.getAttribute('nonce');
```
to
```ts
nonce = nonceElement.nonce
```

seems to fix the issue.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -14,7 +14,7 @@ export let nonce = esmsInitOptions.nonce;
 if (!nonce) {
   const nonceElement = document.querySelector('script[nonce]');
   if (nonceElement)
-    nonce = nonceElement.nonce;
+    nonce = nonceElement.nonce || nonceElement.getAttribute('nonce');
 }
 
 export const onerror = globalHook(esmsInitOptions.onerror || noop);
--- a/src/options.js
+++ b/src/options.js
@@ -14,7 +14,7 @@ export let nonce = esmsInitOptions.nonce;
 if (!nonce) {
   const nonceElement = document.querySelector('script[nonce]');
   if (nonceElement)
-    nonce = nonceElement.getAttribute('nonce');
+    nonce = nonceElement.nonce;
 }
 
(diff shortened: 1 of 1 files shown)


src/options.js
```diff
@@ -14,7 +14,7 @@ export let nonce = esmsInitOptions.nonce;
 if (!nonce) {
   const nonceElement = document.querySelector('script[nonce]');
   if (nonceElement)
-    nonce = nonceElement.getAttribute('nonce');
+    nonce = nonceElement.nonce;
```
guybedford, line 17: Perhaps:

```suggestion
    nonce = nonceElement.nonce || nonceElement.getAttribute('nonce');
```

will fix it if `.nonce` wasn't defined in older Firefox versions.

guybedford (commented):
Thanks for the PR and pointing out the issue! Unfortunately it seems to be failing in older Firefox versions. I'm landing a possible correction now, let's see if that fixes the CI.

src/options.js
MathiasWP: That seems like a fair fix, but it still seems like the *Firefox Browser Test (89)* is failing. Do you know the reason why?

guybedford: Hmm, that test was passing before the fix. I'll rerun the CI, perhaps try it locally?

MathiasWP: Seems to work now! When will this fix be deployed?

guybedford:
Released in 1.3.2.

PR #225, opened 2021-11-19
doc: correct minor syntax error in CSS/JSON modules usage example

benfjohnson:
I discovered a small typo while following the instructions for configuring JSON module support (the error message correctly indicated which property was expected)


guybedford:
Thanks for spotting this one.

PR #227, opened 2021-11-30
Use createObjectURL & iframe.src

lewisl9029:
An alternative approach to addressing Weixing Browser compatibility, first attempt here: https://github.com/guybedford/es-module-shims/pull/165

This version uses iframe.src and createObjectURL, which in theory should work anywhere es-module-shims itself does (since it's already used heavily internally), instead of document.write, which results in a lighthouse score hit and scary warning:

image

I don't have easy access to Weixing Browser to test this, so we'll probably need to roll a new release and ask for help from @xxgjzftd to try it out.

### DIFF
--- a/src/features.js
+++ b/src/features.js
@@ -30,7 +30,7 @@ export const featureDetectionPromise = Promise.resolve(supportsDynamicImportChec
       const iframe = document.createElement('iframe');
       iframe.style.display = 'none';
       document.head.appendChild(iframe);
-      iframe.src = URL.createObjectURL(new Blob([`<script type=importmap nonce="${nonce}">{"imports":{"x":"data:text/javascript,"}}<${''}/script><script nonce="${nonce}">import('x').then(()=>1,()=>0).then(v=>parent._$s(v))<${''}/script>`], {type : 'text/html'}))
+      iframe.src = createBlob(`<script type=importmap nonce="${nonce}">{"imports":{"x":"data:text/javascript,"}}<${''}/script><script nonce="${nonce}">import('x').then(()=>1,()=>0).then(v=>parent._$s(v))<${''}/script>`, 'text/html')
     })
   ]);
 });
--- a/src/features.js
+++ b/src/features.js
@@ -30,7 +30,7 @@ export const featureDetectionPromise = Promise.resolve(supportsDynamicImportChec
       const iframe = document.createElement('iframe');
       iframe.style.display = 'none';
       document.head.appendChild(iframe);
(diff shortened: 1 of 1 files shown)


guybedford:
Seems like this is failing CI? I'm all for this approach if it can pass!

lewisl9029:
> Seems like this is failing CI? I'm all for this approach if it can pass!

Silly mistake on my side haha, should be fixed here: https://github.com/guybedford/es-module-shims/pull/227/commits/4e4af56ac0493c7a31789c7c888c022e92568e15

🤞 for the tests

src/features.js
```diff
@@ -30,8 +30,7 @@ export const featureDetectionPromise = Promise.resolve(supportsDynamicImportChec
       const iframe = document.createElement('iframe');
       iframe.style.display = 'none';
       document.head.appendChild(iframe);
-      // we use document.write here because eg Weixin built-in browser doesn't support setting srcdoc
-      iframe.contentWindow.document.write(`<script type=importmap nonce="${nonce}">{"imports":{"x":"data:text/javascript,"}}<${''}/script><script nonce="${nonce}">import('x').then(()=>1,()=>0).then(v=>parent._$s(v))<${''}/script>`);
+      iframe.src = URL.createObjectURL(new Blob([`<script type=importmap nonce="${nonce}">{"imports":{"x":"data:text/javascript,"}}<${''}/script><script nonce="${nonce}">import('x').then(()=>1,()=>0).then(v=>parent._$s(v))<${''}/script>`], {type : 'text/html'}))
```
guybedford, line 33: We also already have a helper for this that can be imported - https://github.com/guybedford/es-module-shims/blob/main/src/common.js#L5.

guybedford (changes requested):
All passing, good to land just see the one comment which might save a couple more bytes.

guybedford:
Ah actually it is failing in Chrome unfortunately it seems.

lewisl9029:
> Ah actually it is failing in Chrome unfortunately it seems.

I'm hoping this is a flake, all I've changed in that commit was remove the `self.`, which really shouldn't change anything. 🤞 for the latest version with the helper.

src/features.js
lewisl9029: Of course we do 🤦 
Updated here: https://github.com/guybedford/es-module-shims/pull/227/commits/84e792f4283807e0309a1fec43cae433ebd5bcea

lewisl9029:
Looks like we're good as far as tests are concerned!

guybedford (approved):
Perfect, seems like that did it. Thank you!

guybedford:
It turns out this approach does have an effect on Android webviews whereby the `isMainFrame` property is now being triggered as if this were a navigation, whereas before it wasn't. This may have some security / visibility consequences, but for now it doesn't seem to be blocking or worthy of reverting. Let's keep an eye on it though.

guybedford:
@LarsDenBakker I'm happy to revert this PR, we just need to work out what a cheap feature detection might be for weixin browser here. Does anyone have access to this browser to research further?

Issue #232, opened 2021-12-07
esmsInitOptions is not defined


ffortier:
When using esms-options script with a custom resolver hook, the code fails with the following error:

```
err ReferenceError: esmsInitOptions is not defined
    at resolve (es-module-shims.js:322)
    at importShim (es-module-shims.js:443)
```

My script tag looks like this

```
<script type="esms-options">{"shimMode":true,"resolve":"__customResolve"}</script>

<script>
window.__customResolve = (id, parent, defaultResolve) => defaultResolve(id, parent); // More logic in real life
</script>
```

Live code: https://jsfiddle.net/vujx2ws7/1/

The error is here: https://github.com/guybedford/es-module-shims/blob/main/src/es-module-shims.js#L48

Even if the options.js file parses the `script[type=esms-options]` correctly, the `resolve` expect the options to be defined as a global variable.

A possible workaround is to use that global variable.

PR #233, opened 2021-12-07
Fix/issue 232 esms options

ffortier:
Fixes https://github.com/guybedford/es-module-shims/issues/232

When using a script tag to set the `esms-options`, it was failing with `esmsInitOptions` not defined. The code was using a global variable. This PR uses the `resolveHook` exported by the `options.js` instead.


guybedford:
Thanks for the PR, released in 1.3.6.

PR #237, opened 2021-12-15
feat: onpolyfill note default

guybedford:
This updates the `onpolyfill` hook to apply when first executing polyfilled code instead of just on init, so when there is a real native code path that failed that the polyfill is kicking in to polyfill.

Then, the default hook contains a `console.info` message noting that the native error which will be present can be ignored.

Here's how that looks in Firefox:

image

Hopefully this will help avoid any production confusions.


guybedford:
I've updated the note to try and be clearer:

image

PR #240, opened 2021-12-16
Fix default value of `enforceIntegrity` option documented in README

rmacklin:
This was misleading because the option doesn't default to `true`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,7 +431,7 @@ window.esmsInitOptions = {
   resolve: (id, parentUrl, resolve) => resolve(id, parentUrl), // default is spec resolution
   fetch: (url, options) => fetch(url, options), // default is native
   revokeBlobURLs: true, // default false
-  enforceIntegrity: true, // default true
+  enforceIntegrity: true, // default false
 }
 </script>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -431,7 +431,7 @@ window.esmsInitOptions = {
   resolve: (id, parentUrl, resolve) => resolve(id, parentUrl), // default is spec resolution
   fetch: (url, options) => fetch(url, options), // default is native
   revokeBlobURLs: true, // default false
-  enforceIntegrity: true, // default true
+  enforceIntegrity: true, // default false
```
rmacklin, line 434: I think there's actually no default value; i.e. it's `undefined` if not explicitly set, but that's also true for `revokeBlobURLs` which is documented as `default false` above, so I just followed that example. Since this is used as a boolean option, it seems fine.

guybedford: While internally we don't explicitly set the boolean to save bytes, the option is only binary anyway so I think we're good!

guybedford:
Thanks for the correction here @rmacklin.

rmacklin:
No problem, thank you for all your work on this project!

Issue #228, opened 2021-12-02
Source maps are not allowed to be dynamically loaded in Safari


dhh:
In Safari, and not Chrome or Firefox, source maps are not being loaded when going through the shim. Here are the errors:

<img width="927" alt="Screen Shot 2021-12-02 at 14 47 45" src="https://user-images.githubusercontent.com/2741/144434287-c8fcb834-c22b-430b-891b-705c70dd266e.png">

Here's the source map reference from one of the files:

<img width="502" alt="Screen Shot 2021-12-02 at 14 48 29" src="https://user-images.githubusercontent.com/2741/144434370-a8e55c0e-3b53-4b8b-8aea-9c93545403a0.png">

It loads correctly when you access it directly via https://peaceful-ravine-89231.herokuapp.com/assets/turbo.min.js-12ab1481b54e9f84d4561ac0e6e5328811dc5b233b7cba05444fcdee85912e46.map.

Tested on Safari 15.1 / M1.

guybedford:
Good to know - I believe the issue here is the trailing `//!\n;` means the sourceMappingURL isn't picked up by the normalization step that is supposed to avoid this.

So if you try rewriting the original source to not include those lines it should work.

I could update the sourceMappingURL regex to support arbitrary trailing line comments or empty statements... let's just verify the fix works first?

dhh:
Ah, lemme verify whether that's the issue!

dhh:
Confirmed that this is the problem. When there's not extra stuff below the sourceMappingURL line, Safari is happy. Problem is that for a bunch of complicated reasons we need those extra lines in Rails 7.

guybedford:
Fix released in 1.3.4 - let me know if there are any further issues.

dhh:
Unfortunately that did not fix it. Still see the same issue with 1.3.4.

<img width="924" alt="Screen Shot 2021-12-03 at 09 58 29" src="https://user-images.githubusercontent.com/2741/144574609-ad65bbdc-d10a-47f3-aaa9-c4febd97b410.png">

dhh:
Bingo. That solved it. Thanks a bunch! ✌️

PR #241, opened 2021-12-16
Fix typo and move comment about source map compatibility above the line it documents

rmacklin:
This is a minor follow-up to 76b18ba75273b2bd1330a9b418b100e686d06707 from https://github.com/guybedford/es-module-shims/pull/229 - that refactoring ended up separating the comment from the RegExp it was annotating. I've brought them back together and also fixed a typo: "Ruby 7" was used instead of "Ruby on Rails 7" (the former doesn't exist: [Ruby is currently on version 3](https://www.ruby-lang.org/en/downloads/releases/)). I also extended the comment with a link to the associated issue (https://github.com/guybedford/es-module-shims/issues/228) since I thought it could be helpful for anyone looking for additional context, but let me know if you don't want that.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -277,7 +277,8 @@ function resolveDeps (load, seen) {
   load.S = undefined;
 }
 
-// ; and // trailer support added for Ruby 7 source maps compatibility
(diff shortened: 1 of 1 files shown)


guybedford:
Looks good 👍

PR #246, opened 2022-01-24
fix: optional catch binding for chrome 64

thepassle:
Chrome 64 doesn't handle the absent catch binding, since it's such a minimal code change I was hoping to fix it in the source. Happy to hear your thoughts 🙂


guybedford:
Looks good, sure!

Issue #254, opened 2022-02-08
Option to allow overriding existing import map entries in shim mode


lewisl9029:
Hi again! 🙂 

In an earlier PR we added [this check](https://github.com/guybedford/es-module-shims/pull/213/files#diff-a35b52a7f96052ad4f5adb06ea0b0c5e9bd470334360d0ab8e1ce7e9b64144f4R129) to throw when users attempt to override an existing import map entry.

Fast forward to today, and I actually found a nasty edge case in my custom react-refresh workflow that requires overriding existing import map entries to fix. 😅 

Would you be open to adding an option to bypass the check? Maybe in the form of an `allowOverrides` init option that defaults to false and only takes effect in shim mode?

Happy to send over a PR if you're ok with the idea. Thanks!

guybedford:
Hi @lewisl9029, agreed this could be a useful option to add. If you're able to work on a PR that would be great. Please just try to keep the byte size of the implementation as small as possible!

lewisl9029:
@guybedford random idea related to your comment on byte size: wdty of potentially having a separate build for shim vs polyfill mode to further optimize byte size? It seems unlikely for a site to need to switch between two modes, and even more unlikely that it would ever work.

Might be as straightforward as replacing the shimMode export to a build-time injected boolean and reconfiguring the bundler. Will be a breaking change for users depending on automatic shim-mode detection, but I think it might be worthwhile longer term.

Happy to explore this as a separate follow up PR if that sounds like something worth exploring!

guybedford:
@lewisl9029 if there could be a significant file saving that could be interesting, although I'm not sure it would be such a big saving as most of the code paths are shared. Separating features might well be a byte size reduction approach in future though eg having extensions loaded in based on what is needed. At the same time I quite like that it's one thing at the moment, does simplify the integration process too.

lewisl9029:
Sounds good, I think I'll give it a shot in a quick and dirty way to explore how much we could potentially save, then we could decide if it's worthwhile once we have some actual numbers.

lewisl9029:
PR is ready for review here: https://github.com/guybedford/es-module-shims/pull/257

Also opened a draft to test out the separate builds idea: https://github.com/guybedford/es-module-shims/pull/258

TL;DR: you were right, the savings were pretty insignificant (<0.5kb). 😅

guybedford:
Landed in #257.

Issue #255, opened 2022-02-08
Dynamic import of CSS modules incorrect argument replacement


joelmoss:
The following code:

```html
<script async src="https://ga.jspm.io/npm:es-module-shims@1.4.4/dist/es-module-shims.js"></script>
<script type="module-shim">
  const path = '/application.css'
  await import(path, { assert: { type: 'css' } })
  await import('/application.css', { assert: { type: 'css' } })
</script>
```

Results in:

```javascript
const path = '/application.css'
await importShim(path, { assert: { type: 'css' } }, 'https://mysite.test/authentication/new')
await importShim('/application.css', 'https://mysite.test/authentication/new', { assert: { type: 'css' } })
```

As you can see, when a variable is used as the first argument to `import()`, the arguments to `importShim()` are incorrect (`path, assertion, parentUrl`), instead of the the correct (`path, parentUrl, assertion`)

guybedford:
It seems like this might have been introduced in https://github.com/guybedford/es-module-shims/pull/248/files.

Seems it's just a simple rewrite ordering issue, thanks for posting.

guybedford:
//cc @ICodeMyOwnLife

ICodeMyOwnLife:
I’ll look into it. Never used dynamic import with dynamic paths before :)

ICodeMyOwnLife:
@guybedford 
With this source code
```js
... import(dynamicJsonPath, { assert: { type: "json" } }).then( ...
```
`lexer` will parse into something like this
```js
{
  a: -1
  d: 794
  e: 846
  n: undefined
  s: 801
  se: 846
  ss: 794
}
```
Note that `e` is equal to `se` and `source.slice(s, e)` is evaluated to be `dynamicJsonPath, { assert: { type: "json" } }`. With just these parameters we can't extract only the first argument. We may need to update `lexer` to give us more information?

guybedford:
@ICodeMyOwnLife it could definitely be useful to track the `,` position in es-module-lexer, but the easier option for now might be to just change the `importShim` function to rather have the signature:

```js
importShim(specifier: string, opts: object, parent?: string)
importShim(specifier: string, parent?: string)
```

where a `typeof arg` for the second argument could branch on being a string, undefined or object. Let me know if that sounds like it can work to you.

ICodeMyOwnLife:
It’s a good idea, I’ll try.

PR #257, opened 2022-02-10
Allow importmap overrides in shim mode with option

lewisl9029:
As discussed in #254, this PR adds a `override` boolean option to allow overriding of existing import map entries in shim mode, defaulting to false to preserve current behavior. Overall increase in footprint is 7 bytes to 9167 according to the `npm run footprint` command.

A few hoops I had to jump through:

- Depending on ./config.js in ./common.js introduced a circular dependency, which caused some weird issues in tests. So I ended up refactoring the noop function into its own module to untangle it. I also tried inlining the noop function everywhere, but it ended up making the bundle larger than using a separate module.

- For testing, I ended up adding a separate module for this option because I couldn't figure out a way to make the test coexist in the same module without conflicting with the existing importmap overrides tests, which require this option to be off. Open to alternatives!


README.md
```diff
@@ -432,6 +433,7 @@ window.esmsInitOptions = {
   fetch: (url, options) => fetch(url, options), // default is native
   revokeBlobURLs: true, // default false
   enforceIntegrity: true, // default false
+  override: true, // default false
```
guybedford, line 436: Let's give it a more descriptive name, perhaps something like `mapOverrides: true`?

src/common.js
```diff
@@ -113,7 +113,7 @@ export function resolveUrl (relUrl, parentUrl) {
 function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap) {
   for (let p in packages) {
     const resolvedLhs = resolveIfNotPlainOrUrl(p, baseUrl) || p;
-    if (outPackages[resolvedLhs] && (outPackages[resolvedLhs] !== packages[resolvedLhs])) {
+    if ((!shimMode || !override) && outPackages[resolvedLhs] && (outPackages[resolvedLhs] !== packages[resolvedLhs])) {
```
guybedford, line 116: Rather than moving the "noop" to a separate file, what about creating a separate `resolve.js` module?

guybedford (approved):
Looks great, thanks for your work on this!

README.md
lewisl9029: Yep, sounds good, went a bit overboard with the bytes saving here. 😅

src/common.js
lewisl9029: Good call, let me give it a shot!

lewisl9029: Addressed here: https://github.com/guybedford/es-module-shims/pull/257/commits/8140a862d64fa661533b6f4ea2367044bef8724d

README.md
lewisl9029: Addressed here: https://github.com/guybedford/es-module-shims/pull/257/commits/39432375f5d59f61271f1d5bf8b0360322c4a54d

lewisl9029:
With the latest changes we're only 2 bytes higher at 9162 now! I guess refactoring out the resolve functions must have saved enough bytes to make up for everything else, somehow.

guybedford:
Thanks for the quick turnaround on this one - footprint is great!

Will release soon, just hoping to get https://github.com/guybedford/es-module-shims/issues/255 fix in as well.

PR #260, opened 2022-02-12
fix: #255 importShim getting parentUrl

ICodeMyOwnLife:
This tries to fix https://github.com/guybedford/es-module-shims/issues/255

1. Update `resolveDeps` so that dynamic imports will be converted the form `importShim(id[, assertion][, parentUrl])`

2. Update `importShim` to expect `parentUrl` if present will be the last argument


guybedford:
Thanks, looks good.

PR #265, opened 2022-02-23
fix: alternative 227 implementation

guybedford:
This implements an alternative to https://github.com/guybedford/es-module-shims/pull/227 avoiding the navigation event.

// cc @LarsDenBakker 

@lewisl9029 I'm not easily able to test this on Weixin browser, but I assume the approach should work there in it being a `document.write` restriction and not a `contentDocument` restriction.

This PR also includes the build system upgrade to Chomp.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Use Chomp
         run: cargo install chompbuild
       - name: Chomp Test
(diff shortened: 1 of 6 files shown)


LarsDenBakker:
That's great! Thanks for the fast implementation

guybedford:
No problem - released in 1.4.7!

PR #269, opened 2022-03-08
Hot reloading support via import.meta.hot

guybedford:
This implements the `import.meta.hot` API from Vite (https://vitejs.dev/guide/api-hmr.html).

The hot reloader is implemented as an extension using initialized hooks that executes before es-module-shims.

In addition this PR adds support for `nativePassthrough: false` to disable native passthrough, which is set automatically when enabling hot reloading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,11 @@ For the remaining ~4% of users, the highly performant (see [benchmarks](#benchma
 The following modules features are polyfilled:
 
 * [Import Maps](#import-maps) polyfill.
-* [JSON](#json-modules) and [CSS modules](#css-modules) with import assertions when enabled.
-* [Wasm modules](#wasm-modules) with support for Source Phase Imports when enabled.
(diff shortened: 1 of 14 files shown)


abhishiv:
Hey @guybedford 

Just found this PR. Are you still planning on merging it?

https://github.com/guybedford/es-module-shims/issues/380#issuecomment-1657476630
> Thanks for sharing. There is actually a hot reload implementation for this project in https://github.com/guybedford/es-module-shims/pull/269, I just never landed it (not sure why... it works fine!). The main issue is just that it needs integration with a server that can send the reload events, but Chomp does support it as well.

If you are hesitant on merging it because of this reason, maybe I suggest instead of baking in websocket support, a function should be exposed which anyone can call in order to process these messages.

I'm interested in this, because i'm building a browser based IDE at @gratico, and hence I think a this approach is more generic and useful than making it dependent on websocket.

PR #290, opened 2022-05-11
Add nonce to feature detection iframe

iainbeeston:
es-module-shims creates an iframe to do feature
detection. However, if a site is using CSP then
this iframe is blocked unless the CSP config allows
frame-src to use blobs, which prevents es-module-shims
from running. In chrome that error looks like:

    Refused to frame 'blob:http://localhost:3000/477a88e6-fb4c-48fd-a677-2ba7fb69fc45'
    because it violates the following Content Security Policy directive:
    "default-src 'self' https:". Note that 'frame-src' was not explicitly set,
    so 'default-src' is used as a fallback.

In safari it looks like:

    Refused to load blob:https://cors-test.herokuapp.com/12cecc46-329e-4791-a73b-4824852d1ad1
    because it appears in neither the child-src directive nor the default-src
    directive of the Content Security Policy.

To prevent this, I've updated `feature.js` to set a nonce
value on the iframe, and this fixes the error.

### DIFF
--- a/src/features.js
+++ b/src/features.js
@@ -28,6 +28,7 @@ export const featureDetectionPromise = Promise.resolve(supportsDynamicImportChec
       };
       const iframe = document.createElement('iframe');
       iframe.style.display = 'none';
+      iframe.setAttribute('nonce', nonce);
       iframe.srcdoc = `<script type=importmap nonce="${nonce}">{"imports":{"x":"data:text/javascript,"}}<${''}/script><script nonce="${nonce}">import('x').then(()=>1,()=>0).then(v=>parent._$s(v))<${''}/script>`;
       document.head.appendChild(iframe);
     })


iainbeeston:
This PR is part of my attempt to fix https://github.com/rails/rails/issues/45057

guybedford:
Thanks for the PR, happy to merge. Any ideas why the existing CSP tests aren't catching this?

iainbeeston:
@guybedford I've investigated why the csp tests weren't catching this and I've discovered a few things:

1. I'm using v1.3.4 which has this bug and adding the nonce fixes it
2. The reason the tests don't pick this up is because there is no `default-src` rule defined in the csp `<meta>` tag and so blobs are allowed with any source (I'd recommend adding `default-src: 'none'` or maybe `default-src: 'self'` as that will detect this kind of error in future)
3. This fix is actually unnecessary in the latest version of es-module-shims because it no longer uses a blob for the `<iframe>` source created in `features.js`, and so the `script-src` rule defined in the `<meta>` works for the `<script>` tag in the `<iframe>`

Based on that I'd suggest reverting this PR but adding `default-src: 'none'` to `test-css.html`. What do you think?

iainbeeston:
The change from blob to script was in #265, which means the original issue should have been fixed from v1.4.7 onwards

iainbeeston:
I've added #293 to update `test-csp.html`

PR #293, opened 2022-05-16
Set default csp policy

iainbeeston:
Previously there was no default set so
any files could be loaded from anywhere
(except for scripts which were allow-listed).

This changes the csp so that by default
only localhost can be used. This should
detect issues like #290.


guybedford:
Thanks for tracking this one down, looks great thanks.

Issue #295, opened 2022-05-26
ESM self-import: different behaviors between shim and browser native modes


evisong:
Hi, thanks for this great project that enables us to use import-maps in production.

Recently we run into a tricky issue. When creating ESM bundles using latest Webpack 5, there's some self-import code at the bottom of entry ESM file. The bundles work well in latest Chrome, however they fail to initialize when handled in shim mode.

I've made a minimal example to reproduce this. Below is the JS part, the complete example is uploaded to https://gist.github.com/evisong/0e70d3ee5215d751e7d57d8378de9c96.

```js
export const b = {
  m: () => console.log('Run m')
};

import { b as c } from './module-self-import.mjs';

const m = c.m;

export default function a() {
  console.log('Run a');
  m();
}
```

In native mode, the browser console shows:

```
Run inline
Run a
Run m
```

However in shim mode, the browser console prompts an error:

```
module.mjs:7 Uncaught TypeError: Cannot read properties of undefined (reading 'm')
    at module.mjs:7:13
```

To workaround this blocking issue, we have to disable shim mode, so we temporarily inline es module URLs instead of using elegant importmaps.

I'm not sure if this is a common use case since I've only seen it in Webpack generated bundles. Could you share some guidance? Thanks.

guybedford:
Thanks for posting this case, it is useful to know that this is an important feature for Webpack output.

I was able to prototype an approach that can solve this issue in https://github.com/guybedford/es-module-shims/pull/296.

Unfortunately it does require some lexer work in https://github.com/guybedford/es-module-lexer/issues/116 to properly get it to a point where it can be comprehensive and something we can merge.

I don't currently have much bandwidth to work on the above further, but will leave the approach open to track progress. If anyone is interested in working on or sponsoring this work further that would be a huge help too.

evisong:
@guybedford Appreciate your quick response. For now, it seems magic to me that the 3 LOC in #296 fixes this. I'll take some time this week to dig into it, after that if I might fully understand it and am confident enough, then I'll be glad to contribute.

tommie:
I stumbled on this issue with import map polyfill in Firefox: https://gist.github.com/tommie/f20e4c297cf86044681abd04df3d6ce5

The #296 fix works for this repro case as well. Thanks!

[...]

Issue #271, opened 2022-03-14
Enhancement suggestion: Provide cyclical live bindings support


badde-media:
Since the polyfill doesn't provide live bindings support, problems with cycles in your code may occur. The workaround you've shown in [issue 270](https://github.com/guybedford/es-module-shims/issues/270) could easily be automated, so that less experienced devs wouldn't run into such problems, no?

guybedford:
Note this is specifically live bindings support for cycle shell modules - live bindings are fully supported otherwise.

Unfortunately there is no easy way to make this work, this is explicitly listed in the readme as a special case of the polyfill technique.

The best we could likely do is try to detect errors in cycle modules and add additional context for the error message.

guybedford:
Closing as a cantfix / wontfix.

guybedford:
Note, I finally came up with an approach that can possibly finally solve this issue in https://github.com/guybedford/es-module-shims/pull/296 but it still would require some more work to be a proper fix.

PR #296, opened 2022-05-27
wip: progressive cycle case prototype

guybedford:
This implements an approach for progressive cycle handling, solving https://github.com/guybedford/es-module-shims/issues/295 and https://github.com/guybedford/es-module-shims/issues/271.

This approach cannot be merged in its current form and is just a prototype, but is a proof of concept that demonstrates a viable approach for solving these problems.

A mergeable non-buggy approach would require https://github.com/guybedford/es-module-lexer/issues/116 to be implemented in es-module-lexer. This PR can track that work in progress.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -246,7 +246,7 @@ function resolveDeps (load, seen) {
   for (const dep of load.d)
     resolveDeps(dep, seen);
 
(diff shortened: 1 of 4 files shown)


guybedford:
Full implementation at https://github.com/guybedford/es-module-shims/pull/311.

Issue #297, opened 2022-06-02
Possible React Native timing issues


guybedford:
Per https://github.com/rails/importmap-rails/pull/128, it's possible that the changes in https://github.com/guybedford/es-module-shims/pull/287 may have caused a new timing issue for React native apps with dom event retriggering.

There is a current replication in https://github.com/crohr/test-importmap but this requires a comprehensive Rails & React Native setup to test.

More investigation work would be needed to isolate / track down any issue here further.

PR #298, opened 2022-06-03
fix: iOS webview support

guybedford:
This fixes https://github.com/guybedford/es-module-shims/issues/297, where React native iOS webviews were not supporting the feature detection.

This effectively reverts https://github.com/guybedford/es-module-shims/pull/265.

Now that we have https://github.com/guybedford/es-module-shims/pull/292, the original concern in https://github.com/guybedford/es-module-shims/pull/227 is no longer as important since `document.write` will only ever be used in Safari so that the warning should be much less prominent now, and even less so when Safari supports `HTMLScriptElement.supports` (which it likely would do along with import maps I assume).

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -451,7 +451,7 @@ function getOrCreateLoad (url, fetchOpts, parent, source) {
   load.L = load.f.then(async () => {
     let childFetchOpts = fetchOpts;
     load.d = (await Promise.all(load.a[0].map(async ({ n, d }) => {
-      if (d >= 0 && !supportsDynamicImport || d === 2 && !supportsImportMeta)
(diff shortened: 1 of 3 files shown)


src/es-module-shims.js
```diff
@@ -451,7 +451,7 @@ function getOrCreateLoad (url, fetchOpts, parent, source) {
   load.L = load.f.then(async () => {
     let childFetchOpts = fetchOpts;
     load.d = (await Promise.all(load.a[0].map(async ({ n, d }) => {
-      if (d >= 0 && !supportsDynamicImport || d === 2 && !supportsImportMeta)
+      if (d >= 0 && !supportsDynamicImport || d === -2 && !supportsImportMeta)
```
guybedford, line 454: (This is an unrelated typo)

PR #300, opened 2022-06-11
Enable es-module-shims usage in web workers

costingeana:
The purpose of this commit is to enable the usage of `es-module-shims` in module web workers, which at the moment are supported in Chrome based browsers and Safari > 15. As you know, in Firefox the support is still missing.

However, web workers cannot access objects like `document` or `window`. Therefore, I had to do the following changes in `es-modules-shims` code:

- add some simple pieces of code that check whether `document/window` exists before doing something with the `document/window` object; it's not elegant, but these checks are necessary;
- add a function that computes the `baseUrl`. In fact, this function was already defined in `es-module-shims` version `0.4.7`.
- I also added a new function `setImportMap` which can be used to set the import map from outside (obviously). This function is useful when you start a web worker which uses `es-module-shims`.

It’s important to mention that in web workers, I use the `es-module-shims.wasm.js` output. Why? Because of the `dynamicImportScript` function; I’m not sure if there is a solution for the CSP version in web workers.
Also, I want to emphasize that I didn't add new tests because no new functionality was added. The existing tests cover the changes from this pull request. The usage of `es-module-shims` in web worker is a decision that belongs to the client code.

Below, I attach a piece of code that starts a web worker which uses `es-module-shims`:
```
// baseURL, esModuleShimsURL, aURL are provided

const workerScriptUrl = URL.createObjectURL(new Blob(
        [`importScripts('${new URL(esModuleShimsURL, baseURL).href}');importShim.setImportMap(${JSON.stringify(importShim.getImportMap())});importShim('${new URL(aURL, baseURL).href}').catch(e=>setTimeout(()=>{throw e}))`],
        { type: 'application/javascript' }))

const worker = new Worker(workerScriptUrl);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,10 +292,21 @@ While in polyfill mode the same restrictions apply that multiple import maps, im
 To make it easy to keep track of import map state, es-module-shims provides a `importShim.getImportMap` utility function, available only in shim mode.
 
 ```js
-const importMap = importShim.getImportMap()
+const importMap = importShim.getImportMap();
 
 // importMap will be an object in the same shape as the json in a importmap script
 ```
+
+#### Setting current import map state
+To make it easy to set the import map state, es-module-shims provides a `importShim.setImportMap` utility function, available only in shim mode.
+
+```js
+// importMap will be an object in the same shape as the json in a importmap script
+const importMap = { imports: {/*...*/}, scopes: {/*...*/} };
+
+importShim.setImportMap(importMap);
+```
+
 
 ### Dynamic Import
 
@@ -435,16 +446,21 @@ An example of ES Module Shims usage in web workers is provided below:
  */
 function getWorkerScriptURL(aURL) {
   // baseURL, esModuleShimsURL are considered to be known in advance
-  // esModuleShimsURL - must point to the non-CSP build of ES Module Shims, namely the `es-module-shim.wasm.js` output: es-module-shims/dist/es-module-shims.wasm.js
+  // esModuleShimsURL - must point to the non-CSP build of ES Module Shims, 
+  // namely the `es-module-shim.wasm.js` output: es-module-shims/dist/es-module-shims.wasm.js
 
   return URL.createObjectURL(new Blob(
-    [`importScripts('${new URL(esModuleShimsURL, baseURL).href}');importShim.setImportMap(${JSON.stringify(importShim.getImportMap())});importShim('${new URL(aURL, baseURL).href}').catch(e=>setTimeout(()=>{throw e}))`],
+    [
+      `importScripts('${new URL(esModuleShimsURL, baseURL).href}');
+      importShim.setImportMap(${JSON.stringify(importShim.getImportMap())});
+      importShim('${new URL(aURL, baseURL).href}').catch(e => setTimeout(() => { throw e; }))`
+    ],
     { type: 'application/javascript' }))
 }
 
 const worker = new Worker(getWorkerScriptURL('myEsModule.js'));
 ```
-> For now, in web workers must be used the non-CSP build of ES Module Shims, namely the `es-module-shim.wasm.js` output.
+> For now, in web workers must be used the non-CSP build of ES Module Shims, namely the `es-module-shim.wasm.js` output: es-module-shims/dist/es-module-shims.wasm.js.
 
 ## Init Options
 
--- a/README.md
+++ b/README.md
@@ -298,13 +298,13 @@ const importMap = importShim.getImportMap();
 ```
 
 #### Setting current import map state
-To make it easy to set the import map state, es-module-shims provides a `importShim.extendImportMap` utility function, available only in shim mode.
+To make it easy to set the import map state, es-module-shims provides a `importShim.addImportMap` utility function, available only in shim mode.
 
 ```js
 // importMap will be an object in the same shape as the json in a importmap script
 const importMap = { imports: {/*...*/}, scopes: {/*...*/} };
 
-importShim.extendImportMap(importMap);
+importShim.addImportMap(importMap);
 ```
 
 
@@ -452,7 +452,7 @@ function getWorkerScriptURL(aURL) {
   return URL.createObjectURL(new Blob(
     [
       `importScripts('${new URL(esModuleShimsURL, baseURL).href}');
-      importShim.setImportMap(${JSON.stringify(importShim.getImportMap())});
+      importShim.addImportMap(${JSON.stringify(importShim.getImportMap())});
       importShim('${new URL(aURL, baseURL).href}').catch(e => setTimeout(() => { throw e; }))`
     ],
     { type: 'application/javascript' }))
--- a/README.md
+++ b/README.md
@@ -423,6 +423,29 @@ var resolvedUrl = import.meta.resolve('dep', 'https://site.com/another/scope');
 
 Node.js also implements a similar API, although it's in the process of shifting to a synchronous resolver.
 
+### Module Workers
+ES Module Shims can be used in module workers in browsers that provide dynamic import in worker environments, which at the moment are Chrome(80+), Edge(80+), Safari(15+).
+
+An example of ES Module Shims usage in web workers is provided below:
+```js
+/**
+ * 
+ * @param {string} aURL a string representing the URL of the module script the worker will execute.
+ * @returns {string} The string representing the URL of the script the worker will execute.
+ */
+function getWorkerScriptURL(aURL) {
+  // baseURL, esModuleShimsURL are considered to be known in advance
+  // esModuleShimsURL - must point to the non-CSP build of ES Module Shims, namely the `es-module-shim.wasm.js` output: es-module-shims/dist/es-module-shims.wasm.js
+
+  return URL.createObjectURL(new Blob(
+    [`importScripts('${new URL(esModuleShimsURL, baseURL).href}');importShim.setImportMap(${JSON.stringify(importShim.getImportMap())});importShim('${new URL(aURL, baseURL).href}').catch(e=>setTimeout(()=>{throw e}))`],
+    { type: 'application/javascript' }))
+}
+
+const worker = new Worker(getWorkerScriptURL('myEsModule.js'));
+```
+> For now, in web workers must be used the non-CSP build of ES Module Shims, namely the `es-module-shim.wasm.js` output.
+
 ## Init Options
 
 Provide a `esmsInitOptions` on the global scope before `es-module-shims` is loaded to configure various aspects of the module loading process:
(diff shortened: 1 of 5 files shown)


src/es-module-shims.js
```diff
@@ -502,14 +509,16 @@ function domContentLoadedCheck () {
...
 // this should always trigger because we assume es-module-shims is itself a domcontentloaded requirement
-document.addEventListener('DOMContentLoaded', async () => {
-  await initPromise;
-  domContentLoadedCheck();
-  if (shimMode || !baselinePassthrough) {
-    processImportMaps();
-    processScriptsAndPreloads();
-  }
-});
+if(hasDocument) {
```
guybedford, line 512:
```suggestion
if (hasDocument) {
```

src/es-module-shims.js
```diff
@@ -468,13 +471,17 @@ function getOrCreateLoad (url, fetchOpts, parent, source) {
 }
 
 function processScriptsAndPreloads () {
+  if (!hasDocument) return;
```
guybedford, line 474: Rather than having this guard here, let's rather move it to the caller side.

src/es-module-shims.js
```diff
@@ -468,13 +471,17 @@ function getOrCreateLoad (url, fetchOpts, parent, source) {
...
+
   for (const script of document.querySelectorAll(shimMode ? 'script[type=module-shim]' : 'script[type=module]'))
     processScript(script);
   for (const link of document.querySelectorAll(shimMode ? 'link[rel=modulepreload-shim]' : 'link[rel=modulepreload]'))
     processPreload(link);
 }
 
 function processImportMaps () {
+  if (!hasDocument) return;
```
guybedford, line 483: Same as above, let's rather guard the caller.

src/env.js
```diff
@@ -46,15 +48,31 @@ export function setShimMode () {
...
+  if (!baseUrl && typeof location !== 'undefined') {
+    baseUrl = location.href.split('#')[0].split('?')[0];
+    const lastSepIndex = baseUrl.lastIndexOf('/');
+    if (lastSepIndex !== -1)
+      baseUrl = baseUrl.slice(0, lastSepIndex + 1);
+  }
+
+  return baseUrl;
+}
+export const baseUrl = getBaseURL();
```
guybedford, line 67: How about something like:

```suggestion
export const baseUrl = hasDocument ? document.baseURI : `${location.protocol}//${location.host}${location.pathname.includes('/') ? location.pathname.slice(0, location.pathname.lastIndexOf('/') + 1) : location.pathname}`;
```

Since it won't work if there is no `location` anyway.

guybedford (commented):
Thanks this looks good!

Could you add a small documentation section for workers, explaining how this can be used? This may well be useful to others if they know about it.

See note about unifying some of the guards.

Perhaps worker CSP may be possible using import scripts of a blob? Not sure, but may well make sense to keep current CSP separation just unsupported for now.

costingeana:
Hello @guybedford and sorry for the delay.

I applied the suggested changes and I added a small documentation section; however, I'm not sure if it's in the right place.
For now, CSP is not supported for workers. I will try your suggestion in the next days. 

Thank you.

src/es-module-shims.js
```diff
@@ -97,6 +100,7 @@ function metaResolve (id, parentUrl = this.url) {
 
 importShim.resolve = resolveSync;
 importShim.getImportMap = () => JSON.parse(JSON.stringify(importMap));
+importShim.setImportMap = (importMapIn) => Object.assign(importMap, JSON.parse(JSON.stringify(importMapIn)));
```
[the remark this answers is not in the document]
guybedford:
```suggestion
importShim.setImportMap = importMapIn => {
  if (!shimMode) throw new Error('Unsupported in polyfill mode.');
  importMap = JSON.parse(JSON.stringify(importMapIn));
}
```

guybedford (commented):
Couple more doc and code comments, but I think we're there. Thanks I really like how simple this approach turned out.

I'll aim to merge in the next few days.

src/es-module-shims.js
```diff
@@ -135,41 +139,42 @@ const initPromise = featureDetectionPromise.then(() => {
...
-    new MutationObserver(mutations => {
-      for (const mutation of mutations) {
-        if (mutation.type !== 'childList') continue;
-        for (const node of mutation.addedNodes) {
-          if (node.tagName === 'SCRIPT') {
-            if (node.type === (shimMode ? 'module-shim' : 'module'))
-              processScript(node);
-            if (node.type === (shimMode ? 'importmap-shim' : 'importmap'))
-              processImportMap(node);
+    if (hasDocument) {
```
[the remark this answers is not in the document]
costingeana: I'm not sure if they can be joined because at line 179 there is a `return lexer.init;` Maybe I'm wrong, but I think that line should be executed even if there is no `document` object (e.g. inside the worker  scope).

src/es-module-shims.js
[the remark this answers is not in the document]
costingeana: I used Object.assign in order to preserve the existing `importMap` object defined as `let importMap = { imports: {}, scopes: {} };` (in fact to preserve the reference to the object, in case someone was referencing it). But I think this assumption is futile. I will change as you suggested.

src/es-module-shims.js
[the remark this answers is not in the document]
guybedford: We should probably just always return `lexer.init` actually, completely unguarded, there's no need to stop that since it's hardly blocking anything and always being done anyway.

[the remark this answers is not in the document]
guybedford: Yes, exactly!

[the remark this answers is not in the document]
costingeana: All right. Thanks!

src/es-module-shims.js
```diff
@@ -97,6 +100,10 @@ function metaResolve (id, parentUrl = this.url) {
 
 importShim.resolve = resolveSync;
 importShim.getImportMap = () => JSON.parse(JSON.stringify(importMap));
+importShim.setImportMap = importMapIn => {
+  if (!shimMode) throw new Error('Unsupported in polyfill mode.');
+  importMap = JSON.parse(JSON.stringify(importMapIn));
```
guybedford, line 105: Of course I completely forgot we need to properly normalize the import map, in which case we should probably make this an `extendImportMap` function instead of `setImportMap` as well:

```suggestion
  importMap = resolveAndComposeImportMap(importMapIn, pageBaseUrl, importMap);
```

src/es-module-shims.js
```diff
@@ -97,6 +100,10 @@ function metaResolve (id, parentUrl = this.url) {
 
 importShim.resolve = resolveSync;
 importShim.getImportMap = () => JSON.parse(JSON.stringify(importMap));
+importShim.setImportMap = importMapIn => {
```
guybedford, line 103:
```suggestion
importShim.extendImportMap = importMapIn => {
```

README.md
```diff
@@ -292,11 +292,22 @@ While in polyfill mode the same restrictions apply that multiple import maps, im
...
 ```
 
+#### Setting current import map state
+To make it easy to set the import map state, es-module-shims provides a `importShim.setImportMap` utility function, available only in shim mode.
+
+```js
+// importMap will be an object in the same shape as the json in a importmap script
+const importMap = { imports: {/*...*/}, scopes: {/*...*/} };
+
+importShim.setImportMap(importMap);
```
guybedford, line 307:
```suggestion
importShim.extendImportMap(importMap);
```

README.md
```diff
@@ -292,11 +292,22 @@ While in polyfill mode the same restrictions apply that multiple import maps, im
...
 
 ```js
-const importMap = importShim.getImportMap()
+const importMap = importShim.getImportMap();
 
 // importMap will be an object in the same shape as the json in a importmap script
 ```
 
+#### Setting current import map state
+To make it easy to set the import map state, es-module-shims provides a `importShim.setImportMap` utility function, available only in shim mode.
```
guybedford, line 301:
```suggestion
To make it easy to set the import map state, es-module-shims provides a `importShim.extendImportMap` utility function, available only in shim mode.
```

guybedford:
I added the `extendImportMap` reimplementation and rename, please just check that still works out correctly for your use cases.

README.md
```diff
@@ -423,6 +434,34 @@ var resolvedUrl = import.meta.resolve('dep', 'https://site.com/another/scope');
...
+ */
+function getWorkerScriptURL(aURL) {
+  // baseURL, esModuleShimsURL are considered to be known in advance
+  // esModuleShimsURL - must point to the non-CSP build of ES Module Shims, 
+  // namely the `es-module-shim.wasm.js` output: es-module-shims/dist/es-module-shims.wasm.js
+
+  return URL.createObjectURL(new Blob(
+    [
+      `importScripts('${new URL(esModuleShimsURL, baseURL).href}');
+      importShim.setImportMap(${JSON.stringify(importShim.getImportMap())});
```
guybedford, line 455:
```suggestion
      importShim.extendImportMap(${JSON.stringify(importShim.getImportMap())});
```

README.md
```diff
@@ -292,11 +292,22 @@ While in polyfill mode the same restrictions apply that multiple import maps, im
...
 
 ```js
-const importMap = importShim.getImportMap()
+const importMap = importShim.getImportMap();
 
 // importMap will be an object in the same shape as the json in a importmap script
 ```
 
+#### Setting current import map state
+To make it easy to set the import map state, es-module-shims provides a `importShim.extendImportMap` utility function, available only in shim mode.
```
guybedford, line 301:
```suggestion
To make it easy to set the import map state, es-module-shims provides a `importShim.addImportMap` utility function, available only in shim mode.
```

README.md
```diff
@@ -292,11 +292,22 @@ While in polyfill mode the same restrictions apply that multiple import maps, im
...
+
+```js
+// importMap will be an object in the same shape as the json in a importmap script
+const importMap = { imports: {/*...*/}, scopes: {/*...*/} };
+
+importShim.extendImportMap(importMap);
```
guybedford, line 307:
```suggestion
importShim.addImportMap(importMap);
```

README.md
guybedford, line 455:
```suggestion
      importShim.addImportMap(${JSON.stringify(importShim.getImportMap())});
```

src/es-module-shims.js
```diff
@@ -97,6 +100,10 @@ function metaResolve (id, parentUrl = this.url) {
 
 importShim.resolve = resolveSync;
 importShim.getImportMap = () => JSON.parse(JSON.stringify(importMap));
+importShim.extendImportMap = importMapIn => {
```
guybedford, line 103:
```suggestion
importShim.addImportMap = importMapIn => {
```

PR #301, opened 2022-06-16
Improve TypeScript types

blixt:
Updating the TypeScript types to include recent additions. Also attempts to accurately represent the types in the `registry` map, though that could use some extra scrutiny.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+import type { parse } from "es-module-lexer";
+
 interface ESMSInitOptions {
   /**
    * Enable Shim Mode
@@ -18,7 +20,7 @@ interface ESMSInitOptions {
   /**
    * Disable retriggering of document readystate
    */
-  noLoadEventRetriggers: true,
+  noLoadEventRetriggers: true;
 
   /**
    * #### Skip Processing Stability
@@ -45,24 +47,28 @@ interface ESMSInitOptions {
 
   /**
    * #### Error hook
-   * 
+   *
    * Register a callback for any ES Module Shims module errors.
-   * 
+   *
    */
   onerror: (e: any) => any;
 
   /**
    * #### Resolve Hook
-   * 
+   *
    * Only supported in Shim Mode.
-   * 
+   *
    * Provide a custom resolver function.
    */
-  resolve: (id: string, parentUrl: string, resolve: (id: string, parentUrl: string) => string) => string | Promise<string>;
+  resolve: (
+    id: string,
+    parentUrl: string,
+    resolve: (id: string, parentUrl: string) => string
+  ) => string | Promise<string>;
 
   /**
    * #### Fetch Hook
-   * 
+   *
    * Only supported in Shim Mode.
    *
    * > Stability: Non-spec feature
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -126,14 +132,53 @@ interface ESMSInitOptions {
...
+  b: string | undefined;
+  /** Shell URL */
+  s: string | undefined;
+  /** Needs shim */
+  n: boolean;
+  /**
+   * Module type
+   * @deprecated This field is never set.
+   */
+  t: null;
```
blixt, line 177: I guess this property could be removed completely because I don't see it being used (there is another `t` property being returned by the fetch functionality, but that object does not appear to ever overlap with the one set in `registry`).

index.d.ts
```diff
@@ -1,3 +1,5 @@
+import type { parse } from "es-module-lexer";
```
blixt, line 1: I just realized that this may require `es-module-lexer` to become a non-dev dependency 🤔 If that's the case, then maybe we should just simplify the type down to `[imports: unknown[], exports: unknown[], facade: boolean]`?

guybedford:
The resolver / `importShim` types look great. For the registry unfortunately that's a private API (so won't be providing semver-level backwards-compat), so I'd be weary to type it unless TypeScript has some explicit way of indicating something is a private / non semver API.

guybedford:
The resolver / `importShim` types look great. For the registry unfortunately that's a private API (so won't be providing semver-level backwards-compat), so I'd be weary to type it unless TypeScript has some explicit way of indicating something is a private / non semver API.

blixt:
We can exclude the `_r` property and leave that up to the intrepid developer that needs to use it to type in their private repo (which is what we're doing today at Framer). It would be nice to get some level of long-term support for a reverse lookup of blob URL to import specifier though. The following code is how I'm using the `_r` property today, maybe it could be exposed in a follow-up PR?

```typescript
/**
 * Make blob URLs (modules shimmed locally to work with our dynamic import map)
 * human readable again, very useful for debugging purposes.
 */
function replaceBlobURLs(value: string): string {
    const reverseMap = new Map<string, string>()

    // Replace blob URLs with import map entries.
    const importMap = importShim.getImportMap()
    for (const [specifier, url] of Object.entries(importMap.imports)) {
        if (!url.startsWith("blob:")) continue
        reverseMap.set(url, specifier)
    }

    // Replace blob URLs with URL imports.
    for (const [specifier, entry] of Object.entries(importShim._r)) {
        if (!entry.b) continue
        if (reverseMap.has(entry.b)) {
            // Don't override import map entries which should be more accurate.
            continue
        }
        // Sometimes blob URLs point to other blob URLs, so attempt to resolve.
        // TODO: This should probably also follow multi-step chains of blob URLs.
        reverseMap.set(entry.b, reverseMap.get(specifier) ?? specifier)
    }

    return value.replace(/blob:https?:\/\/[^/]+\/[a-z0-9-]+/g, blobURL => reverseMap.get(blobURL) ?? blobURL)
}
```

index.d.ts
```diff
@@ -45,24 +45,28 @@ interface ESMSInitOptions {
...
-   * 
+   *
    * Provide a custom resolver function.
    */
-  resolve: (id: string, parentUrl: string, resolve: (id: string, parentUrl: string) => string) => string | Promise<string>;
+  resolve: (
+    id: string,
+    parentUrl: string,
+    resolve: (id: string, parentUrl: string) => string
+  ) => string | Promise<string>;
```
blixt, line 65: Prettier did this but I figured I'd leave it in since it's less likely to cause more diffs in the future (assuming most devs have Prettier running).

guybedford:
Ah makes sense. There may be places with rethrown errors where es-module-shims itself might want to do a reverse lookup. If that becomes the case, we could expose the function internally as well then.

PR #302, opened 2022-06-19
feat: simplify feature detections csp policies

guybedford:
This fixes CSP support with a number of enhancements:

* Fixing the case where `HTMLScriptElement.supports('importmap') === false` skips the import map detection as it should have been doing already
* Updating the internal feature detections to use blobs instead of data URIs so that it's only necessary to add `blob:` to the policy and not both `blob:` and `data:` for easier usage
* The document.write path in the iframe feature detection is again modified to attempt a srcdoc approach first before falling back to document.write. This avoids a bunch of warnings and possible issues with using document.write by default. Hopefully this finally gets us to a middle ground on that case.
* A new clear error message is thrown by the Wasm build when eval is not supported, to avoid possible obscure errors.

### DIFF
--- a/src/dynamic-import.js
+++ b/src/dynamic-import.js
@@ -7,9 +7,10 @@ try {
   eval('');
 }
 catch (e) {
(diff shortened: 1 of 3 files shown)


PonomareVlad:
Does not initialize in safari 15.6 🥲

Снимок экрана 2022-06-20 в 08 25 28

Снимок экрана 2022-06-20 в 08 26 28

PR #311, opened 2022-07-27
fix: self imports.

### LINKED ISSUE #295: ESM self-import: different behaviors between shim and browser native modes

tommie:
Upgrades to es-module-lexer 1.0.0. For returning exports as ExportSpecifier[].

Based on @guybedford's #296.

Closes #295.

### DIFF
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -267,17 +267,15 @@ function resolveDeps (load, seen) {
         if (cycleShell) {
           // circular shell creation
           if (!(blobUrl = depLoad.s)) {
-            function dotOrBracketAccessor(expt) {
-              const ch = depLoad.S[expt.s];
-              return (ch === '"' || ch === "'") ? `[${depLoad.S.slice(expt.s, expt.e)}]` : '.' + expt.n;
-            }
-            const shellExports = depLoad.a[1].map(expt => expt.ln ? expt : {...expt, ln: 'd$_'});
             blobUrl = depLoad.s = createBlob(`export function u$_(m){${
-              shellExports.map(expt => `${expt.ln}=m${dotOrBracketAccessor(expt)}`).join(',')
+              depLoad.a[1].map(({ s, e }, i) => {
+                const q = depLoad.S[s] === '"' || depLoad.S[s] === "'";
+                return `e$_${i}=m${q ? `[` : '.'}${depLoad.S.slice(s, e)}${q ? `]` : ''}`;
+              }).join(',')
             }}${
-              shellExports.map(expt => `let ${expt.ln};`).join('')
+              depLoad.a[1].length ? `let ${depLoad.a[1].map((_, i) => `e$_${i}`).join(',')};` : ''
             }export {${
-              shellExports.map(expt => `${expt.ln} as ${depLoad.S.slice(expt.s, expt.e)}`).join(',')
+              depLoad.a[1].map(({ s, e }, i) => `e$_${i} as ${depLoad.S.slice(s, e)}`).join(',')
             }}\n//# sourceURL=${depLoad.r}?cycle`);
           }
         }
@@ -323,6 +321,7 @@ function resolveDeps (load, seen) {
   if (!hasSourceURL)
     resolvedSource += '\n//# sourceURL=' + load.r;
 
+  console.log(resolvedSource);
   load.b = lastLoad = createBlob(resolvedSource);
   load.S = undefined;
 }
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -321,7 +321,6 @@ function resolveDeps (load, seen) {
   if (!hasSourceURL)
     resolvedSource += '\n//# sourceURL=' + load.r;
 
-  console.log(resolvedSource);
   load.b = lastLoad = createBlob(resolvedSource);
   load.S = undefined;
 }
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -267,15 +267,23 @@ function resolveDeps (load, seen) {
         if (cycleShell) {
           // circular shell creation
           if (!(blobUrl = depLoad.s)) {
+            function dotOrBracketAccessor(expt) {
+              const ch = depLoad.S[expt.s];
+              return (ch === '"' || ch === "'") ? `[${depLoad.S.slice(expt.s, expt.e)}]` : '.' + expt.n;
+            }
             blobUrl = depLoad.s = createBlob(`export function u$_(m){${
               depLoad.a[1].map(
-                expt => expt.n === 'default' ? `d$_=m.default` : `${expt.n}=m.${expt.n}`
+                expt => expt.ln ? `${expt.ln}=m${dotOrBracketAccessor(expt)}` : `d$_=m.default`
               ).join(',')
             }}${
-              depLoad.a[1].map(expt =>
-                expt.n === 'default' ? `let d$_;export{d$_ as default}` : `export let ${expt.n}`
-              ).join(';')
-            }\n//# sourceURL=${depLoad.r}?cycle`);
+              depLoad.a[1].map(
+                expt => expt.ln ? `let ${expt.ln};` : 'let d$_;'
+              ).join('')
+            }export {${
+              depLoad.a[1].map(
+                expt => expt.ln ? `${expt.ln} as ${depLoad.S.slice(expt.s, expt.e)}` : `d$_ as default`
+              ).join(',')
+            }}\n//# sourceURL=${depLoad.r}?cycle`);
           }
         }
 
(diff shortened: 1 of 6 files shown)


src/es-module-shims.js
```diff
@@ -306,6 +306,12 @@ function resolveDeps (load, seen) {
       }
     }
 
+    // support progressive cycle binding updates
+    if (load.s) {
+      const fields = exports.filter(expt => expt.n !== 'default' && expt.ln).map(expt => `${expt.n}: ${expt.ln}`);
```
guybedford, line 311: Rather than `${expt.n}:` we should use the same form as the original exported identifier - `${source.slice(expt.s, expt.e)}` to also support string exports. We also don't actually need a default filter here since `expt.ln` is always guaranteed to be a valid identifier now:

```suggestion
      const fields = exports.filter(expt => expt.ln).map(expt => `${source.slice(expt.s, expt.e)}: ${expt.ln}`);
```

src/es-module-shims.js
```diff
@@ -269,11 +269,11 @@ function resolveDeps (load, seen) {
...
             blobUrl = depLoad.s = createBlob(`export function u$_(m){${
               depLoad.a[1].map(
-                name => name === 'default' ? `d$_=m.default` : `${name}=m.${name}`
+                expt => expt.n === 'default' ? `d$_=m.default` : `${expt.n}=m.${expt.n}`
               ).join(',')
             }}${
-              depLoad.a[1].map(name =>
-                name === 'default' ? `let d$_;export{d$_ as default}` : `export let ${name}`
+              depLoad.a[1].map(expt =>
+                expt.n === 'default' ? `let d$_;export{d$_ as default}` : `export let ${expt.n}`
```
guybedford, line 276: I wonder if we can update this to properly support string exports now / non identifier exports in general rather than just default?

One way would be to number all identifiers (`e$1`, `e$2`, ...) and then export a bag - `export { e$1 as ${source.slice(expt.s, expt.e)}, ... }` kind of thing. Let me know if that makes sense?

guybedford (commented):
Thanks for the update, will be great to land this!

src/es-module-shims.js
tommie: Nice!

src/es-module-shims.js
tommie: Done. We already have the local name, so I don't think we need to generate identifiers.

The two cyclic test cases now create:

```js
export function u$_(m){f=m.f,d$_=m.default}let f;let d$_;export {f as f,d$_ as default}
//# sourceURL=http://localhost:8088/test/fixtures/test-circular1.js?cycle

export function u$_(m){b=m.b,c=m.d,e=m['e f'],d$_=m.default}let b;let c;let e;let d$_;export {b as b,c as d,e as 'e f',d$_ as default}
//# sourceURL=http://localhost:8088/test/fixtures/test-self-import.js?cycle
```

src/es-module-shims.js
```diff
@@ -267,15 +267,23 @@ function resolveDeps (load, seen) {
...
-                name => name === 'default' ? `d$_=m.default` : `${name}=m.${name}`
+                expt => expt.ln ? `${expt.ln}=m${dotOrBracketAccessor(expt)}` : `d$_=m.default`
               ).join(',')
             }}${
-              depLoad.a[1].map(name =>
-                name === 'default' ? `let d$_;export{d$_ as default}` : `export let ${name}`
-              ).join(';')
-            }\n//# sourceURL=${depLoad.r}?cycle`);
+              depLoad.a[1].map(
+                expt => expt.ln ? `let ${expt.ln};` : 'let d$_;'
```
guybedford, line 280: I think we should just skip bindings entirely if there is no local name, as in we can filter them out first.

src/es-module-shims.js
```diff
@@ -267,15 +267,23 @@ function resolveDeps (load, seen) {
...
+              ).join('')
+            }export {${
+              depLoad.a[1].map(
+                expt => expt.ln ? `${expt.ln} as ${depLoad.S.slice(expt.s, expt.e)}` : `d$_ as default`
```
guybedford, line 284: Default should work out naturally, we shouldn't need to branch on it anymore?

guybedford (commented):
Getting there, see last comments.

guybedford:
This seems to be failing on Chrome 695079. I'd advise trying to download that version from https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Win_x64%2F695079%2Fchrome-win.zip?alt=media to test locally why its failing!

guybedford:
Firefox 67 also seems to have an issue.

tommie:
Since this is only failing on some browsers, I propose #313 to have some feedback in the logs.

There also seems to be some infrastructure errors comingled:

```
Spawning browser: /opt/hostedtoolcache/firefox/67.0/x64/firefox -headless
Retrying...
Spawning browser: /opt/hostedtoolcache/firefox/67.0/x64/firefox -headless
node:events:505
      throw er; // Unhandled 'error' event
      ^

Error: spawn /opt/hostedtoolcache/firefox/67.0/x64/firefox ENOENT
      at Process.ChildProcess._handle.onexit (node:internal/child_process:283:19)
...
```

It's in `onexit`, so it's probably trying to wait twice on the old browser process. Or something. I don't know if Karma provides more feedback on these types of issues. Maybe.

src/es-module-shims.js
tommie: I don't understand how. We still need to export `default`, but it doesn't have an `ln`, right?

Moved the `expr.ln` conditional to a `map()` ahead of the template expansion.

src/es-module-shims.js
tommie: Why don't we have to proxy-export the default (which doesn't have a local name)?

src/es-module-shims.js
guybedford: This is just an _early_ export process to help with during-execution accesses - the main cycle update function is called after execution with the full namespace, which would have the default and reexported bindings as well.

src/es-module-shims.js
```diff
@@ -267,15 +267,18 @@ function resolveDeps (load, seen) {
         if (cycleShell) {
           // circular shell creation
           if (!(blobUrl = depLoad.s)) {
+            function dotOrBracketAccessor(expt) {
+              const ch = depLoad.S[expt.s];
+              return (ch === '"' || ch === "'") ? `[${depLoad.S.slice(expt.s, expt.e)}]` : '.' + expt.n;
+            }
+            const shellExports = depLoad.a[1].map(expt => expt.ln ? expt : {...expt, ln: 'd$_'});
```
guybedford, line 274: For early default updates, we should instead explicitly detect `expt.n === 'default'`.

guybedford (commented):
Getting there! Note that nothing will beat the debugging experience of actually running the failing browsers to fix the tests!

src/es-module-shims.js
guybedford: I still don't see how we would update the default without a local name though... I'm pretty sure it's a moot point.

tommie:
> Getting there! Note that nothing will beat the debugging experience of actually running the failing browsers to fix the tests!

Well, if I can avoid messing with browser installations, I will. I'm testing this locally on Chrome 103, FWIW.

* Chrome 69 (?) fails with `Fail Circular dependencies should support shell update import interleaving: Unexpected string`
* Firefox 67 says `Fail Circular dependencies should support shell update import interleaving: missing export name`

Those look like real errors, so perhaps it is time to download some older versions.

src/es-module-shims.js
guybedford: I went ahead and added the ordered binding approach discussed previously!

guybedford:
The remaining issue seems to be that quoted exports aren't supported in older browsers. I think the best approach for now is just to leave these cases out of the tests since this isn't polyfillable.

tommie:
Cool. Thank you so much for maintaining this project, and finishing the PR!

I just wanted import maps to work in Firefox, and being able to simply import a polyfill is awesome.

guybedford:
Thanks again for working on this one, has been a huge help.

PR #312, opened 2022-07-27
Tests can run on any HTTP port, and auto-allocate a port.

tommie:
I have something already running on port 8080, so being able to say `WATCH_MODE=true CI_PORT=0 node test/server.mjs test-shim` is practical.

The default is still port 8080.

Also narrowing down the test server to only localhost instead of all addresses.

Found while working on #311.

### DIFF
--- a/test/server.mjs
+++ b/test/server.mjs
@@ -7,7 +7,7 @@ import open from "open";
 import kleur from 'kleur';
 import { spawn } from 'child_process';
 
-const port = 8080;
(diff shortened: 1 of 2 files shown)


tommie:
Ah, I moved `spawnPs` from global scope, and only tested with `WATCH_MODE=true`, where it's not used in `/done`.

tommie:
#314 to fix the now broken retries. Sorry about that.

PR #326, opened 2022-08-15
compat: dynamic import baseline update

guybedford:
This updates the polyfill baseline from Chrome 61+, Firefox 60+, Safari 10.1+ and Edge 10+ to Chrome 63+, Firefox 67+ and Safari 11.1+, corresponding to a new dynamic import baseline - https://caniuse.com/es6-module-dynamic-import.

The reason for this is that the dynamic import polyfill cannot polyfill top-level await since load events do not wait for the TLA promise completion, therefore to properly support top-level await we need to deprecate the dynamic import polyfill.

The compat loss is only 0.2% of users and is being implemented as a minor release. If there are problems with this type of baseline shift in future we can be more careful about these updates, but for the sake of major version continuity it feels like the right decision in this case given the importance of top-level await.


guybedford:
Closing, to rather make as a major at a later point. Will work on a workaround for TLA for now.

PR #328, opened 2022-08-15
fix: reinstate DOMContentLoaded script attachments

guybedford:
This reinstates a DOM iteration for DOMContentLoaded in case any script injections were missed by the mutation observers / initialization process.


guybedford:
Closing as unnecessary for now, at least until it is shown this is needed.

PR #338, opened 2022-09-22
feat: skip array option

guybedford:
Adds a new array value type to the `skip` option, corresponding to a list of exact URLs to skip or URL paths to skip. Can be relative or absolute URLs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,7 +626,7 @@ In such a case, this double event firing can be disabled with the `noLoadEventRe
 
 When loading modules that you know will only use baseline modules features, it is possible to set a rule to explicitly opt-out modules from rewriting. This improves performance because those modules then do not need to be processed or transformed at all, so that only local application code is handled and not library code.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -626,12 +626,23 @@ In such a case, this double event firing can be disabled with the `noLoadEventRe
 
 When loading modules that you know will only use baseline modules features, it is possible to set a rule to explicitly opt-out modules from rewriting. This improves performance because those modules then do not need to be processed or transformed at all, so that only local application code is handled and not library code.
 
-This can be configured by providing a URL regular expression for the `skip` option:
+The `skip` option supports a string regular expression, function or array of exact module URLs to check:
```
rmacklin, line 629: Is this documentation up to date? It seems like the PR is adding support for arrays, but this line implies functions are also supported.
```suggestion
The `skip` option supports a string regular expression or array of exact module URLs to check:
```

guybedford: Thanks for spotting that.

PR #339, opened 2022-10-08
Fix: New skip implementation was incorrectly called

kevindumanoir:
Moreover, skip code for Regexp was incorrect (s and r variables were inverted)

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -55,7 +55,7 @@ if (Array.isArray(skip)) {
 }
 else if (typeof skip === 'string') {
   const r = new RegExp(skip);
-  skip = s => s.test(r);
+  skip = s => r.test(s);
 }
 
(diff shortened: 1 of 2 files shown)


src/env.js
```diff
@@ -55,7 +55,7 @@ if (Array.isArray(skip)) {
 }
 else if (typeof skip === 'string') {
   const r = new RegExp(skip);
-  skip = s => s.test(r);
+  skip = s => s.match(r);
```
guybedford: Would still prefer test to avoid allocating the match object unnecessarily:

```suggestion
  skip = s => t.test(s);
```

guybedford (approved):
Thanks!

src/env.js
kevindumanoir: Done!

PR #347, opened 2023-01-04
Update Firefox import maps compatibility w/o shims

laurinneff:
See https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/108#apis


guybedford:
Thanks for the update!

Issue #356, opened 2023-02-21
Not properly recognized on Firefox 109


morgoth:
I'm using the ES module shim by the https://github.com/rails/importmap-rails (which I believe uses this code).

On the Firefox 109, looks like the feature is not properly detected as it's supported natively in this version.
My code:
```js
document.addEventListener('DOMContentLoaded', () => {
  console.log("I run")
})
```
is run twice when loading the shim script.

Please see the details on https://github.com/rails/importmap-rails/issues/175

guybedford:
Which version of es-module-shims are you using? Make sure you are using the latest version 1.5.18 here. The detection is definitely correct, and I couldn't get a double execution in my own test case. Otherwise please share an exact replication for further investigation.

guybedford:
Ah apologies, I'm on slow wifi and had an older version on this machine. Posted a fix in https://github.com/guybedford/es-module-shims/pull/358.

PR #358, opened 2023-02-23
fix: domcontentloaded duplication

guybedford:
Resolves https://github.com/guybedford/es-module-shims/issues/356 in ensuring that under baseline passthrough we always avoid domcontentloaded and readystatechange event retriggering.


morgoth:
@guybedford Thank you. Can we have a new release?

Issue #368, opened 2023-04-05
Hit & miss loading of modules using 1.7.1


tomgreenhill:
Hey Guy, 

Thanks for your previous assistance!

I'm having issues with modules loading intermittently between refreshes. I'm unsure what's causing it as I receive the following log message:

`[Log] ^^ Module TypeError above is polyfilled and can be ignored ^^ (es-module-shims.js, line 1)`

<img width="834" alt="Screenshot 2023-04-05 at 13 53 42" src="https://user-images.githubusercontent.com/1922302/230087071-0407d652-1585-4a6a-8ebc-f6aa9a3612a7.png">

But then sometimes the modules on the page don't load. Am using 1.7.1 and can provide a preview link if required. Happy to assist debugging wherever possible.

Thanks.

guybedford:
Hey @tomgreenhill sure we can look into this, it sounds like it might be something around the load events to me, as opposed to the polyfill engaging this time.

Can you run the the debug build of es-module-shims (https://unpkg.com/browse/es-module-shims@1.7.1/dist/es-module-shims.debug.js), and share the log output / the difference in logging between a successful and unsuccessful run?

Another thing to try is if the `noLoadEventRetriggers` helps or hinders - https://github.com/guybedford/es-module-shims#no-load-event-retriggers.

lewisl9029:
I think I might have a fix at #369, would appreciate some 👀 when you have the chance. 🙏

tomgreenhill:
So sorry I haven't tested this yet. I've been moving  so all over the place. Will try and put some time aside in the next week or so to look over this again. Thanks for the update!

Issue #363, opened 2023-03-17
Polyfill does not always fire on iOS


bakura10:
Hi,

I have tried to migrate our code to using importmap and, under iOS (I tried on 16 and 15) it sometimes work, and sometimes does not (navigating between pages cause the issue):

<img width="540" alt="image" src="https://user-images.githubusercontent.com/1198915/225803635-bc0a8f68-9f0a-48b8-a7ed-c30bbf930ba1.png">

While sometimes, it will work:

<img width="536" alt="image" src="https://user-images.githubusercontent.com/1198915/225803707-bc23aff6-4b42-433e-b865-19c7c101f365.png">

I am not doing anything fancy here (asset_url is a Shopify construct to generate the URL of a given resource, but this happens server side so on the client the complete URL is just rendered):

```
<script async src="{{ 'es-module-shims.min.js' | asset_url }}"></script>

    <script type="importmap">
      {%- comment -%}On Safari 16.3 and lower, a polyfill is used to load importmap{%- endcomment -%}
      {
        "imports": {
          "vendor": "{{ 'vendor.min.js' | asset_url }}",
          "theme": "{{ 'theme.js' | asset_url }}",
          "photoswipe": "{{ 'photoswipe.min.js' | asset_url }}"
        }
      }
    </script>

    <script type="module" src="{{ 'vendor.min.js' | asset_url }}"></script>
    <script type="module" src="{{ 'theme.js' | asset_url }}"></script>
```

What could be happening here that would cause the polyfill to sometime render, sometime not?

Thanks.

EDIT : from what I can see, due to the fact the script is async, it seems to fire up _before_ the other modules, so the TypeError is fired after the polyfill being executed:

<img width="540" alt="image" src="https://user-images.githubusercontent.com/1198915/225804303-8d58a705-d6c5-42a1-9309-0b1220bd2d08.png">

guybedford:
Can you share a replication of the issue that I can test out here? Yes there are a number of different timings that can apply for the polyfill depending on the exact network order and how the page is set up. All of the various cases should be properly handled through.

bakura10:
Hi @guybedford ,

Thanks for your answer :). I don't have a small reproduction case, but you can replicate it here: https://prestige-new-version.myshopify.com/ (to enter you need to enter the password **rewblo**)

[...]

PR #369, opened 2023-04-21
Verify feature detection message

lewisl9029:
I was also running into intermittent feature detection failures similar to the ones mentioned in https://github.com/guybedford/es-module-shims/issues/368 once I upgraded to 1.7.1 on https://reflame.app, so I made a custom build with some additional logging on the message callback function and found a pretty big clue:

Screenshot 2023-04-20 at 4 28 18 PM

Looks like the root of the issue might be that we're not verifying the source of the message, so anything that calls postMessage before es-module-shims (in this case the react dev tools extension) can cause feature detection to fail, often non-deterministically due to races.

Looking at this [previous related issue](https://github.com/guybedford/es-module-shims/issues/363), I think this is the more likely cause for that one too rather than some security policy (i.e. something on the page called postMessage with `undefined` as the message). So I removed the previous branch where we continue to resolve feature detection even after failing validation, and instead opting to return early and wait for the next message. WDYT?

Also, I added the debug build as an exported module so it can be accessed from the npm package as well.

### DIFF
--- a/src/features.js
+++ b/src/features.js
@@ -31,14 +31,14 @@ export let featureDetectionPromise = Promise.resolve(dynamicImportCheck).then(()
     iframe.style.display = 'none';
     iframe.setAttribute('nonce', nonce);
     function cb ({ data }) {
-      // failed feature detection (security policy) -> revert to default assumptions
-      if (Array.isArray(data)) {
-        supportsImportMaps = data[0];
-        supportsImportMeta = data[1];
-        supportsCssAssertions = data[2];
-        supportsJsonAssertions = data[3];
+      const isFeatureDetectionMessage = Array.isArray(data) && data[0] === 'esms'
+      if (!isFeatureDetectionMessage) {
+        return;
       }
-      else if (self.ESMS_DEBUG) console.info(`es-module-shims: feature detection failure, using defaults`);
+      supportsImportMaps = data[1];
+      supportsImportMeta = data[2];
+      supportsCssAssertions = data[3];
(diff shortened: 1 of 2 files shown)


lewisl9029:
Hmm... Looks like we have a test failure for Firefox 67.0. Not sure if it's a flake or a real failure (can't seem to trigger a retry from my side).

src/features.js
```diff
@@ -31,14 +31,14 @@ export let featureDetectionPromise = Promise.resolve(dynamicImportCheck).then(()
     iframe.style.display = 'none';
     iframe.setAttribute('nonce', nonce);
     function cb ({ data }) {
-      // failed feature detection (security policy) -> revert to default assumptions
-      if (Array.isArray(data)) {
-        supportsImportMaps = data[0];
-        supportsImportMeta = data[1];
-        supportsCssAssertions = data[2];
-        supportsJsonAssertions = data[3];
+      const isFeatureDetectionMessage = Array.isArray(data) && data[0] === 'esms'
```
lewisl9029, line 34: We could potentially make this a bit more resistant to forgery by generating some random bytes instead of this hardcoded string. But I don't see what a potential attacker has to gain by taking over feature detection, so there might not be a need?

guybedford: Agreed, we don't need to be forge resistant.

guybedford (approved):
Thanks for working this one out. I tend to agree it seems more likely security blocking would block the messaging entirely, as opposed to passing primitives, so I'm happy to land this under that assumption.

PR #378, opened 2023-07-03
WASM support

pvh:
Here's the start of a WASM support patch that (lest anyone assume I knew what I was doing) was designed & implemented by @guybedford. I did at least adjust it so it worked for me!

From here, I imagine a proper implementation would mostly involve adding some tests. I didn't want this code to get lost so I'm pushing it up as a draft for now in case anyone else is interested.


guybedford (approved):

guybedford:
Thanks for putting it up, would be glad to merge it.

PR #396, opened 2023-11-06
Fix the `skip` option

### LINKED ISSUE #384: mismatch between code and type
https://github.com/guybedford/es-module-shims/blob/1.8.0/index.d.ts#L52

dts say `skip` is `RegExp`.

https://github.com/guybedford/es-module-shims/blob/1.8.0/src/env.js#L51-L59

But the js code treat it like `RegExp | Array<string> | string`

Is it inaccurate type?

Or, is it the wasm implement is using that type?

koddsson:
Fixes #384.

This change starts accepting a regular expression proper as a `skip` option. This seems to have been indicated by the existing types and counter-intuitively the types that are actually accepted aren't represented in the TypeScript declaration.

This pull request adds the `string[] | string` types to the `skip` property in the options type and creates a `skip` wrapper in case the variable passed in is a regular expression.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -49,7 +49,7 @@ interface ESMSInitOptions {
    * By default, this expression supports jspm.dev, dev.jspm.io and
    * cdn.pika.dev.
    */
-  skip?: RegExp;
+  skip?: RegExp | string[] | string
 
   /**
    * #### Error hook
--- a/src/env.js
+++ b/src/env.js
@@ -56,6 +56,8 @@ if (Array.isArray(skip)) {
 else if (typeof skip === 'string') {
   const r = new RegExp(skip);
   skip = s => r.test(s);
+} else if (skip instanceof RegExp) {
+  skip = s => skip.test(s);
 }
 
 const eoop = err => setTimeout(() => { throw err });


koddsson:
Damn the firefox tests seem flaky in CI 😭. Here's the failing build for posterity while I re-run the build: https://github.com/guybedford/es-module-shims/actions/runs/6772254424/job/18404391999?pr=396

koddsson:
Thinking about this more I feel like I should add tests here since this wasn't caught by tests already.

guybedford:
This is great thank you!

Yes, tests would be preferable.

The Firefox test flakes are a really weird thing where the Firefox binary itself gets cleaned up and no longer exists! Haven't figured that one out to date...

guybedford (approved):

guybedford:
Note - I'm waiting on landing this in case you want to add tests.

koddsson:
> Note - I'm waiting on landing this in case you want to add tests.

I do want to add tests but I couldn't figure out the test setup.

Am I on the right tracking thinking I should add something like:

```html
<script type="module-shim">
  import test from "skipped.js";
  order.push(6);
</script>
```

and a `"skip": new RegExp(/skipped.js/)` to the options bag in `test/test-shim.html`. And then I can add a assertion in `test/shim.js` that the library has delegated the load of `skipped.js` to the native module loader? (I haven't figured out how to do that assertion yet).

guybedford:
Yeah agreed this is hard to test. I mean, it clearly works, so we can just land it too if it's too tricky to figure out a loader check.

PR #412, opened 2024-02-28
Updated Firefox stats

marklundin:
It looks like Firefox supports dynamic import in Worker Contexts as of ~113. See [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1540913) and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#browser_compatibility)


guybedford:
Thanks for sharing, we should reintroduce the worker polyfill in this case - https://github.com/guybedford/es-module-shims/pull/17.

marklundin:
Absolutely, that would be great

Issue #419, opened 2024-04-18
Import SHim generates wrong code....


jogibear9988:
I've a ffile wich looks like this:

                        return {
                            element: new mfcSimulation.MfcSimulation(service, plant),
                        };
                    case MfcViews.ActionsDesigner:
                        import('../components/mfc/mfc-actions-designer.js');
                        return {
                            element: new GenericTableViewBuilder(MfcActionCombinedDTO)
                                .setServiceName(service.Name)
                                .setTitle(title)
                                .setTableViewExtension(new MfcActionGridControls())
                                .setDetailsControl(MfcActionsDesigner.is)
                                .setDetailsInTab(true)
                                .build()
                        };
                    case MfcViews.LiveView:
                    .....
                    case WmsViews.ActionsDesigner:
                        import('../components/wms/wms-actions-designer.js');
                        return {
                            element: new GenericTableViewBuilder(BasicActionDTO)
                                .setService(ServiceType.WMSService)
                                .setTitle(title)
                                .setDetailsControl('wms-actions-designer')
                                .setDetailsInTab(true)
                                .build()
                        };
                    case WmsViews.Hardware:

after import shim it looks like this (in chrome source tab):

                        return {
                            element: new mfcSimulation.MfcSimulation(service, plant),
                        };
                    case MfcViews.ActionsDesigner:
                        importShim('../components/mfc/mfc-actions-designer.js', 'http://127.0.0.1:8088/dist/view/viewCreator.js');
                        return {
                            element: new GenericTableViewBuilder(MfcActionCombinedDTO)
                                .setServiceName(service.Name)
                                .setTitle(title)
                                .setTableViewExtension(new MfcActionGridControls())
                                .setDetailsControl(MfcActionsDesigner.is)
                                .setDetailsInTab(true)
                                .build()
                        };
                    case MfcViews.LiveView:
                    .....
                    case WmsViews.ActionsDesigner:
                        importShim(
     ;import{u$_}from'blob:http://127.0.0.1:8088/aecf5795-7472-4cac-9778-e71f4d74817e';try{u$_({ViewCreator:ViewCreator})}catch(_){};
    '../components/wms/wms-actions-designer.js', 'http://127.0.0.1:8088/dist/view/viewCreator.js');
                        return {
                            element: new GenericTableViewBuilder(BasicActionDTO)
                                .setService(ServiceType.WMSService)
                                .setTitle(title)
                                .setDetailsControl('wms-actions-designer')
[...]

PR #421, opened 2024-04-19
fix cycle binding updates for expressions

guybedford:
Resolves https://github.com/guybedford/es-module-shims/issues/419 in ensuring the progressive binding update interleaving for cycles only applies after a declaration import.


jogibear9988:
@guybedford could this be merged and be released?

PR #466, opened 2025-02-13
Make Wasm instantiation synchronous

tomayac:
As per the conversation at the CG meeting. There're more instances that I'm not sure about, but this fixes the `README`, which is where I got confused from.

### DIFF
--- a/README.md
+++ b/README.md
@@ -538,7 +538,7 @@ And for the source phase:
 ```html
 <script type="module">
 import source mod from './app.wasm';
-const instance = await WebAssembly.instantiate(mod, { /* ...imports */ });
+const instance = WebAssembly.Instance(mod, { /* ...imports */ });
 </script>
 ```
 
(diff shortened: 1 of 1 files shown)


guybedford:
I fixed this up to use the `new` keyword in https://github.com/guybedford/es-module-shims/commit/c1b287077ae758264b592d9160e6f44c8b970a80.

tomayac:
> I fixed this up to use the `new` keyword in [c1b2870](https://github.com/guybedford/es-module-shims/commit/c1b287077ae758264b592d9160e6f44c8b970a80).

Duh, sorry, and thank you!

guybedford:
One other issue with this one - I get this error in our test suite on Chrome 133:

```
RangeError: WebAssembly.Instance is disallowed on the main thread, if the buffer size is larger than 8MB. Use WebAssembly.instantiate, or use the flag `--enable-features=WebAssemblyUnlimitedSyncCompilation
```

The problem though here is that `new WebAssembly.instance(module)` should be fine as it's not doing compilation, it's only `new WebAssembly.instance(buffer)` that should give this error.

@tomayac where would be a good place to follow-up on this error further? In the mean time we might still need to recommend the `await WebAssembly.instantiate(mod)` approach as it doesn't give the warning.

tomayac:
> @tomayac where would be a good place to follow-up on this error further? In the mean time we might still need to recommend the `await WebAssembly.instantiate(mod)` approach as it doesn't give the warning.

Seems like you solved this yourself, but for anyone who happens to be following this situation, the relevant bug is https://issues.chromium.org/issues/398265337.

PR #468, opened 2025-02-22
Revert #466 readme change

guybedford:
This switches back to the `await WebAssembly.instantiate(...)` advice due to the 8MB warning message we still get (pending https://issues.chromium.org/issues/398265337).

//cc @tomayac


tomayac:
Oh, that's interesting. Thanks for the heads-up. I've CC:ed myself on the bug.

PR #474, opened 2025-04-22
feat: support import defer polyfill

guybedford:
This adds support for the new `import defer` proposal as a form of syntax stripping in the polyfill.

All we do, is that when the `import-defer` feature is enabled, we replace `import defer * as ns from 'foo'` with `import * as ns from 'foo'`.

This can enable a useful polyfill for defer though:

* Use modern `import defer` syntax for modern browser users with defer support, getting the performance benefits.
* On older browsers without this support, module loading fails on this syntax error.
* This syntax error gets polyfilled via this simple stripping from es-module-shims, allowing the app to fully work on older browsers.

// cc @nicolo-ribaudo for reference

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ The following modules features are polyfilled:
 * [Import Maps](#import-maps) polyfill.
 * [JSON](#json-modules) and [CSS modules](#css-modules) with import assertions when enabled.
 * [Wasm modules](#wasm-modules) with support for Source Phase Imports when enabled.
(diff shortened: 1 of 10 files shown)


guybedford:
Very interesting you're able to get an entire workflow together for this. That seems very useful in development workflows to be able to understand deferral semantics and performance benefits.

es-module-shims is a production polyfill optimized for initialization time and polyfill size. The goal being for it not to run in modern browsers but there as a safety net in older browsers. So I think the defer polyfill here will become useful in that workflow in due course, when defer support is at 80% of browsers and the "polyfill" is there for the 20% compat cases being supported, not the 80% that want the defer performance benefits.

PR #478, opened 2025-04-23
feat: enable json-modules by default

guybedford:
This PR enables JSON Modules by default in es-module-shims, allowing applications to polyfill this syntax without needing it to be explicitly enabled as a feature.

JSON modules have 85% browser support now, so most browsers will get baseline passthrough under this default enabling.

One repercussion though is that since JSON modules aren't implemented in Firefox, applications will always have their graphs analyzed for JSON usage.

To avoid this, it could be worth updating this PR to also support a new `polyfillDisable` option supporting `polyfillDisable: ['json-modules']` so that passthrough can still happen in Firefox when users know the application won't use JSON modules. I'm not sure if the benefits outweigh the costs for that yet though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ If using more modern features like CSS Modules or JSON Modules, these need to be
 
 ```html
 <script>
-window.esmsInitOptions = { polyfillEnable: ['css-modules', 'json-modules', 'wasm-modules', 'typescript'] }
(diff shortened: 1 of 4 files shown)


guybedford:
For now I'll avoid the `polyfillDisable` approach, but we can always reconsider it if it turns out to be needed. I'm hopful being forward looking should work out though for the 80% here.

PR #487, opened 2025-04-26
fix: remove _d override

guybedford:
I don't think we should be writing this global value in a way that can't be changed. I feel bad enough making `window.importShim` unwriteable.

What setting this does, is force the dynamic import check in older versions of ES Module Shims to fail - so that those old versions will think that dynamic import is not supported when in fact it is. This would not disengage the polyfill though it would in fact actually result in the polyfill polyfilling dynamic import and then using a dynamic import shim.

Therefore I cannot see how this would functionally disable an older version of ES Module Shims, short of there being very specific logic around the change in polyfill behaviours when it thinks dynamic import is not supported when it is.

//cc @frandiox I think I may need to revert this one unfortunately.

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -20,8 +20,10 @@ export const optionsScript = hasDocument ? document.querySelector('script[type=e
 export const esmsInitOptions = optionsScript ? JSON.parse(optionsScript.innerHTML) : {};
 Object.assign(esmsInitOptions, self.esmsInitOptions || {});
 
+export const version = self.VERSION;
+
 const r = esmsInitOptions.version;
-if (self.importShim || (r && r !== self.VERSION)) {
+if (self.importShim || (r && r !== version)) {
   if (self.ESMS_DEBUG)
     console.info(
       `es-module-shims: skipping initialization as ${r ? `configured for ${r}` : 'another instance has already registered'}`
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -25,7 +25,8 @@ import {
   hotReload as hotReloadEnabled,
   defaultFetchOpts,
   defineValue,
-  optionsScript
+  optionsScript,
+  version
 } from './env.js';
 import {
   supportsImportMaps,
@@ -146,6 +147,7 @@ importShim.addImportMap = importMapIn => {
   if (!shimMode) throw new Error('Unsupported in polyfill mode.');
   composedImportMap = resolveAndComposeImportMap(importMapIn, pageBaseUrl, composedImportMap);
 };
+importShim.version = version;
 
 const registry = (importShim._r = {});
 // Wasm caches
--- a/src/es-module-shims.js
+++ b/src/es-module-shims.js
@@ -157,7 +157,6 @@ defineValue(self, 'importShim', Object.freeze(importShim));
 const shimModeOptions = { ...esmsInitOptions, shimMode: true };
 if (optionsScript) optionsScript.innerHTML = JSON.stringify(shimModeOptions);
 self.esmsInitOptions = shimModeOptions;
-defineValue(self, '_d', undefined);
 
 const loadAll = async (load, seen) => {
   seen[load.u] = 1;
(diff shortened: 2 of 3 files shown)


frandiox:
Hi 👋  

>What setting this does, is force the dynamic import check in older versions of ES Module Shims to fail - so that those old versions will think that dynamic import is not supported when in fact it is. This would not disengage the polyfill though it would in fact actually result in the polyfill polyfilling dynamic import and then using a dynamic import shim.

Yes, it seems to instruct the older version of `importShim` to use a polyfilled version of the dynamic import, that's correct.

However, since `importShim` is frozen by the version that loads first, the polyfilled version should not really run so it seems to not matter. Polyfilling the dynamic import is basically a "collateral effect" of freezing `_d` to undefined, but it's never used (to my understanding and testing). The main effect is actually setting [`supportsDynamicImport`](https://github.com/guybedford/es-module-shims/blob/2.0.0/src/dynamic-import.js#L16) to `false`, which avoids running the [feature detection code here](https://github.com/guybedford/es-module-shims/blob/2.0.0/src/features.js#L28), which is what seems to really break with double loading. Other parts of the code, like [`processScriptsAndPreloads`](https://github.com/guybedford/es-module-shims/blob/2.0.0/src/es-module-shims.js#L664), which runs on init, won't do anything thanks to switching `shimMode` to `true`, so the polyfilled dynamic import never runs, to my understanding.

This is all assuming that the new version loads first. If it loads later, it shouldn't even do the `_d` stuff because it will realize an older version is already installed, and will let it be.

That was my reasoning but I might have missed something? That said, feel free to remove this if you don't feel comfortable with the change! I originally wanted to create a custom version with these changes but thought they might be interesting for the official build as well. -- Even if this doesn't make it to the official polyfill, I would appreciate any insight you might have about this reasoning, in case it's still useful for my usecase using a custom polyfill build 🙏

guybedford:
> which avoids running the [feature detection code here](https://github.com/guybedford/es-module-shims/blob/2.0.0/src/features.js#L28), which is what seems to really break with double loading

Can you explain why the feature detection code is breaking double loading here? I would have thought that importing a few blobs would be relatively harmless?

frandiox:
>Can you explain why the feature detection code is breaking double loading here? I would have thought that importing a few blobs would be relatively harmless?

I've been checking more into detail and I think I found something interesting. Let's call the new version of the polyfill that gets installed first "NEW", and the older version that will be installed last "OLD" (since the opposite is not a problem due to the new checks added in NEW):

I've reduced the problematic code in the feature detection of the OLD polyfill to [this line](https://github.com/guybedford/es-module-shims/blob/2.0.0/src/features.js#L89), specifically `parent.postMessage(['esms'].concat(a),'*'))`. This is not related to the listener added in this very OLD polyfill, but to the listener added by the NEW polyfill that lands first (which uses the [same event parameter key](https://github.com/guybedford/es-module-shims/blob/main/src/features.js#L46)): it gets the message and disables itself, so we basically end up with 2 polyfills disabled or in a weird state 😰 -- and this is a race condition, it depends on who emits the `esms` event first (since the listener is removed after 1 run).

That's why setting `_d` to undefined => `supportsDynamicImport` in OLD polyfill fixes the issue, it prevents it from sending a message to the NEW polyfill.

The good news are, if we change the `esms` event to any other token in the NEW polyfill / new release (like `esms2` or anything), it seems to fix the event collision and we don't need to use `_d` anymore. In this case, the OLD polyfill would still run its feature detection initially (add message listener, remove it, etc), but at least it wouldn't interfere with the NEW polyfill anymore.

Would that be a sensible compromise? Or does freezing `_d` make more sense to avoid the old polyfill running more code?

guybedford:
@frandiox thanks for diving into this further, that makes complete sense. What I've added here instead is an update to the string in the postMessage so that it explicitly includes the version of the shim in it. This way versions will only see messages meant for themselves.

With this fix, does that resolve the issue with not setting this global?

frandiox:
>With this fix, does that resolve the issue with not setting this global?

Yes, I believe it does, according to the test I did yesterday. Thanks for the changes!

PR #506, opened 2025-07-01
fix: synthetic events are dispatched before native ones

### LINKED ISSUE #504: The synthetic `DOMContentLoaded` and `load` events are dispatched before `document.body` is defined in some situations.
Hi again 👋 

We are seeing a weird issue with the `DOMContentLoaded` (and `load`) event emitted by the polyfill. I haven't been able to reproduce it locally so I'm just writing about it here to see if it rings a bell for you.

The problem is that the synthetic event fires too early when `document.body` is not defined yet, which, to my understanding, means it's fired before the native counterpart. The log stack traces we have point to the polyfill:

```
...callback
[native code] dispatchEvent
.../es-modules-shim.2.4.0.js:1 domContentLoadedCheck
```

We've also seen the same for the `load` event (emitted in `loadCheck` according to the stack trace) but it seems to be much more infrequent than `DOMContentLoaded`.

---

The following is just a shot in the dark to see if we get lucky but, as mentioned, I can't reproduce it locally yet so please take this with a grain of salt. Context:

- We load the polyfill with `async`, which I guess could make the code run before the document is fully parsed for big pages.
- All the errors I've seen are in Safari versions around 17.x and 18.5, which means they already support basic module imports.
- `baselineSupport` is likely `false` because these browsers don't support all the features enabled in 2.4.0 by default (multi-impormaps, etc.)

Considering all of this and reading the polyfill code, we can find something that might be a problem:

- It seems the code checks `m` [here](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L763) to differentiate between direct calls from other parts of the code  (i.e. manual `domContentLoadedCheck()`, where `m = undefined`), and where the call is made by a a listener (i.e. `addEventListener('name', domContentLoadedCheck)`, which should get `m = event`).
- However, in [this line](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L907) it sets the callback as `loadPromise.then(domContentLoadedCheck)`, which sends the resolution of `loadPromise` as its parameter.
- If we then check the resolution of `loadPromise`, we see the promise is created by `topLevelLoad`. There's [this particular line](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L309) that seem to align with all the context mentioned above which returns `undefined` instead of `null` ([like in a few lines above](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L293), which doesn't get to that because there's no baseline support). This could cause running `domContentLoadedCheck` with `undefined` (`shimMode` is falsy, and `load.n` is likely falsy too because there's native support for modules).
- If we check the full flow now, **I think** there's a way where we:
  1. Start `processScript` with `domContentLoadedCnt` set to `1`.
  2. Increment `domContentLoadedCnt` to `2` [here](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L869) (`isDomContentLoadedScript` is truthy).
  3. Run `topLevelLoad` and enqueue `domContentLoadCheck` [here](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L907).
  4. Due to `baseLineSupport` being falsy and having basic module import support, we end up returning undefined [here](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L309)
  5. We reduce `domContentLoadedCnt` to 0 [here](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L766-L768), making the condition truthy to...
  6. Dispatch a synthetic `DOMContentLoaded` event [here](https://github.com/guybedford/es-module-shims/blob/607f1b2bf20299ff379d7838734cea98a5f5522c/src/core.js#L771) before the native event is fired (sometimes).

Does that make sense? 🙏  -- I think the same process works also for the `load` event.
Would returning `null` instead of `undefined` in that line cause other issues? 🤔 

---

Our current workaround is to add the event listener we need for `DOMContentLoaded` to `window` instead of `document`, since the synthetic event doesn't bubble up. This doesn't work for `load`, though, since that one is dispatched on `window`.

frandiox:
Closes #504 

I don't see any JSDoc in the code so I've only added a small comment at the top of the function. Is that OK?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -306,7 +306,7 @@ export const topLevelLoad = async (
         console.info(
           `es-module-shims: early exit after graph analysis of ${url} - graph ran natively without needing polyfill`
         );
-      return;
+      return null;
     }
     if (source) {
       return await dynamicImport(createBlob(source), source);
--- a/src/core.js
+++ b/src/core.js
@@ -264,6 +264,7 @@ let importMapPromise = initPromise;
 let firstPolyfillLoad = true;
 let legacyAcceptingImportMaps = true;
 
+// This must always return either a module namespace or null
 export const topLevelLoad = async (
   url,
   parentUrl,
--- a/src/core.js
+++ b/src/core.js
@@ -263,7 +263,7 @@ let importMapPromise = initPromise;
 let firstPolyfillLoad = true;
 let legacyAcceptingImportMaps = true;
 
-  // This must always return either a module namespace or null
+// This must always return either a module namespace or null
 export async function topLevelLoad(
   url,
   parentUrl,


guybedford:
Specifically - all browsers without baseline passthrough, so ones that need to do feature detections.

guybedford:
I've posted a follow-up PR here in https://github.com/guybedford/es-module-shims/pull/519.

frandiox:
👋 Sorry for the late reply, I've been out for a while. Should we then close this PR since there are other related issues and wait for the refactor?

guybedford:
@frandiox this is very hard for me to prioritise at the moment so I don't expect I'll get to it anytime soon unfortunately.

If you have a PR or fix that you'd like to recommend in the mean time, I'm happy to support it, otherwise feel free to close the PR as well.

frandiox:
Oops, forgot about this one. Going to close it for now since it hasn't come up as a problem again internally. The issue can remain open for docs purposes. Thanks!

PR #509, opened 2025-07-02
support */*+json (ex. json-ld)

easrng:
adds support for json-ld and similar `*/*+json` formats (as per the html spec) and makes this function a few bytes smaller minzipped

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -549,7 +549,7 @@ const initTs = async () => {
   if (!esmsTsTransform) esmsTsTransform = m.transform;
 };
 
-const contentTypeRegEx = /^((text|application)\/(x-)?|[^\/;]+\/[^\/;]*\+)((java|type)script|wasm|json|css)(;|$)/;
+const contentTypeRegEx = /^(text|application)\/([^\/;]*\+|x-)?((java|type)script|wasm|json|css)(;|$)/
 async function defaultSourceHook(url, fetchOpts, parent) {
   let res = await doFetch(url, fetchOpts, parent),
     contentType,
--- a/src/core.js
+++ b/src/core.js
@@ -553,7 +553,7 @@ const contentTypeRegEx = /^(text|application)\/([^\/;]*\+|x-)?((java|type)script
 async function defaultSourceHook(url, fetchOpts, parent) {
   let res = await doFetch(url, fetchOpts, parent),
     contentType,
-    [, , , , type, jsts] = (contentType = res.headers.get('content-type') || '').match(contentTypeRegEx) || [];
+    [, , , type, jsts] = (contentType = res.headers.get('content-type') || '').match(contentTypeRegEx) || [];
   type =
     /\.m?ts([?#]|$)/.test(url) ? 'ts'
     : jsts ? jsts[0] + 's'
--- a/src/core.js
+++ b/src/core.js
@@ -549,13 +549,14 @@ const initTs = async () => {
   if (!esmsTsTransform) esmsTsTransform = m.transform;
 };
 
-const contentTypeRegEx =
-  /^(?:[^/;]+\/(?:[^/+;]+\+)?(json)|(?:text|application)\/(?:x-)?((java|type)script|wasm|css))(?:;|$)/;
 async function defaultSourceHook(url, fetchOpts, parent) {
   let res = await doFetch(url, fetchOpts, parent),
     contentType,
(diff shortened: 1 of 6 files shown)


src/core.js
```diff
@@ -549,24 +549,33 @@ const initTs = async () => {
   if (!esmsTsTransform) esmsTsTransform = m.transform;
 };
 
-const contentTypeRegEx = /^(text|application)\/((x-)?javascript|wasm|json|css|typescript)(;|$)/;
+// https://mimesniff.spec.whatwg.org/#parse-a-mime-type
+// https://mimesniff.spec.whatwg.org/#javascript-mime-type
+// https://mimesniff.spec.whatwg.org/#json-mime-type
+// https://github.com/denoland/deno_media_type/blob/abf2b6808bdff4cd6f01ec2acf51955956956792/src/lib.rs#L235-L239
+const contentTypeRegEx =
+  /^[\n\r\t ]*(?:((?:(?:application\/|text\/)(?:x-)?(?:ecmascript|javascript))|text\/(?:livescript|jscript|javascript1\.[0-5]))|(?:text\/|application\/|[-!~#-'*+.0-9^-z|~]+\/[-!~#-'*+.0-9^-z|~]*\+)(json)|application\/(wasm)|text\/(css)|(application\/(x-)?|text\/)typescript|video\/(mp2t|vnd\.dlna\.mpeg-tts))[\n\r\t ]*(;|$)/;
```
guybedford, line 557: I'm pretty sure the fetch spec will handle leading whitespace trimming for us.

Livescript etc are deprecated. While spec compat is nice, are there real world scenarios here or is this just about matching the spec? I think given the size constraints of the project we must remain on the side of pragmatism in not adding extra bytes unnecessarily.

For typescript, I'd managed to just treat it as a generic fallback without the vnd headers support. Is there a reason we do need to support these as opposed to doing the local extension check which seems mostly equivalent? Or is it for cases where the server masks the extension?

easrng: >I'm pretty sure the fetch spec will handle leading whitespace trimming for us.

oh yeah you're right

> Livescript etc are deprecated. While spec compat is nice, are there real world scenarios here or is this just about matching the spec? I think given the size constraints of the project we must remain on the side of pragmatism in not adding extra bytes unnecessarily.

yeah the js changes were just spec compat, realistically doesn't matter probably. support for `*/*+json` should really be added tho so ex. json-ld (`application/ld+json`) can be imported.

>For typescript, I'd managed to just treat it as a generic fallback without the vnd headers support. Is there a reason we do need to support these as opposed to doing the local extension check which seems mostly equivalent? Or is it for cases where the server masks the extension?

oh yeah there probably isn't any case where a typescript file gets served with video mime types and it doesn't have a .ts extension

probably a good idea to add `x-typescript` support tho, it seems relatively common

guybedford:
I just ran a test in Chrome with the `video/lottie+json` content type and got:

> Failed to load module script: Expected a JSON module script but the server responded with a MIME type of "video/lottie+json". Strict MIME type checking is enforced for module scripts per HTML spec.

So while it's a valid JSON content type, it's not supported per the HTML spec. Supporting `+` is worthwhile though.

See also https://mimesniff.spec.whatwg.org/#json-mime-type.

easrng:
The HTML spec defers to mimesniffing, and it works in firefox

guybedford:
Oh, I missed the very relevant `or` here, in which case this might be a bug in Chrome... will see if I can follow-up.

easrng:
https://wpt.fyi/results/html/semantics/scripting-1/the-script-element/json-module/valid-content-type.html?label=master&label=experimental&aligned&q=json-module it was a chrome bug, it was fixed recently

src/core.js
```diff
@@ -549,24 +549,24 @@ const initTs = async () => {
...
-    if (url.endsWith('.ts') || url.endsWith('.mts')) t = 'ts';
-    else
-      throw Error(
-        `Unsupported Content-Type "${contentType}" loading ${url}${fromParent(parent)}. Modules must be served with a valid MIME type like application/javascript.`
-      );
+  let res = await doFetch(url, fetchOpts, parent),
+    contentType,
+    [, , , type, jsts] = (contentType = res.headers.get('content-type') || '').match(contentTypeRegEx) || [];
+  type =
+    /\.m?ts([?#]|$)/.test(url) ? 'ts'
```
easrng, line 558: Oh whoops this check should be later
On my phone rn so can't push a fix

PR #514, opened 2025-07-02
fix undefined references and lint warnings

easrng:


src/core.js
```diff
@@ -116,7 +115,7 @@ if (shimMode || wasmSourcePhaseEnabled)
     await importMapPromise;
     const url = resolve(id, parentUrl || pageBaseUrl).r;
     const load = getOrCreateLoad(url, defaultFetchOpts, undefined, undefined);
-    if (firstPolyfillLoad && !shimMode && load.n && nativelyLoaded) {
+    if (firstPolyfillLoad && !shimMode && load.n) {
       onpolyfill();
       firstPolyfillLoad = false;
     }
```
guybedford, line 121: Because nativelyLoaded is always false here we can actually remove the entire block so this is a great net code removal to catch actually.

guybedford (approved):

PR #515, opened 2025-07-02
remove internal import cycles

easrng:
the hot reload feature (i think) introduced some import cycles. this takes them back out.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -42,7 +42,7 @@ import {
   featureDetectionPromise
 } from './features.js';
 import * as lexer from '../node_modules/es-module-lexer/dist/lexer.asm.js';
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -917,3 +917,5 @@ const processPreload = link => {
     fetchCache[link.href] = fetchModule(link.href, getFetchOpts(link));
   });
 };
+
+if (hotReloadEnabled) initHotReload(topLevelLoad, importShim);
```
guybedford, line 921: Previously the hooks would be initialized before we got here. Would be worth verifying there is no sync use of hooks here before that. Otherwise this should be at the top of the file with the topLevelLoad as a hoisted function and importShim maybe defined sooner to support that.

PR #517, opened 2025-07-02
move globalThis.importShim definition out of core.js

easrng:
relies on https://github.com/guybedford/es-module-shims/pull/516

moves the global registration code out of core.js so it's easier to add new entrypoints that don't modify the global scope


guybedford:
Seems the `./self.js` import is missing here.

PR #525, opened 2025-07-22
Fix shimMode default inside worker

### LINKED ISSUE #524: Shim mode is no longer correctly determined inside worker
The README claims: "By default, when there is no DOM present, ES Module Shims will switch into shim mode". 

This seems to have been correct up to #269, where the conditional now always checks the init options or (if present) the document queries. Previously this had a ternary defaulting to `true` if no document was present.

fvanderveen:
Fixes #524, by reintroducing the `true` default if there is no document.

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -38,7 +38,7 @@ export const shimMode =
   (hasDocument ?
     document.querySelectorAll('script[type=module-shim],script[type=importmap-shim],link[rel=modulepreload-shim]')
       .length > 0
-  // Without a document, shim mode is always true as we cannot polyfill
+    // Without a document, shim mode is always true as we cannot polyfill
   : true);
 
 export let importHook,


src/env.js
```diff
@@ -35,9 +35,10 @@ if (self.importShim || (r && r !== version)) {
 // shim mode is determined on initialization, no late shim mode
 export const shimMode =
   esmsInitOptions.shimMode ||
-  (hasDocument &&
+  (hasDocument ?
     document.querySelectorAll('script[type=module-shim],script[type=importmap-shim],link[rel=modulepreload-shim]')
-      .length > 0);
+      .length > 0
+  : true);
```
guybedford, line 41:
```suggestion
  // Without a document, shim mode is always true as we cannot polyfill
  : true);
```

guybedford (approved):
Thanks for finding this - just important to add a comment given there is no test so we don't regress again.

guybedford:
Thanks for the ping, yes only have limited time for OSS work here.

PR #528, opened 2025-09-03
fix: catch error for legacy browser

tbvjaos510:
relative https://github.com/guybedford/es-module-shims/pull/246

add catch binding for legacy browsers


guybedford:
Which specific browsers are you looking to support here? We dropped support for most of these browsers previously such that only v1 goes back to polyfilling dynamic import etc. now. So I believe most of these browsers are out of our target now per https://github.com/guybedford/es-module-shims?tab=readme-ov-file#browser-support. But let me know if I'm missing anything.

guybedford:
You're completely right thanks for clarifying - 63 - 66 is in our compat matrix!

Issue #446, opened 2025-01-06
fails with CSP policy that includes require-trusted-types-for 'script'


bman654:
This CSP policy

```
Content-Security-Policy: require-trusted-types-for 'script'; trusted-types: esmoduleshims
```

causes the shim to fail during init:

![image](https://github.com/user-attachments/assets/3f75cfe7-f860-4112-a26b-0b339a888642)

due to the manipulation of some DOM elements via `innerHTML` property.

I *think* the fix is to change the code to something more like:

```js
// let call supply the trusted types name via config.trustedTypesPolicyName to opt-in to trusted types support
// at top of module
const ttpConfig = {
  // normally these methods would santize their input, but we know what we are doing and we trust the
  // input so our methods are merely identity functions
  createHTML: input => input
  createScript: input => input
  createScriptURL: input => input
}

const ttp = (typeof trustedTypes !== "undefined") && config.trustedTypesPolicyName ?
  // we know what we are doing and do not need to sanitize the input
  trustedTypes.createPolicy(config.trustedTypesPolicyName, ttpConfig)
  : ttpConfig;

// anywhere innerHTML is being set
// el.innerHTML = html
el.innerHTML = ttp.createHTML(html)

// anywhere script src is being set
//scriptEl.src = url
scriptEl.src = ttp.createScriptURL(url)

// anywhere Function constructor is being used:
// f = new Function ("a", "b", code)
f = new Function("a", "b", ttp.createScript(code))
```

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for

guybedford:
This proposal does sound like it could work to me as a way to support trusted types. I would be happy to take a PR.

PR #536, opened 2025-12-17
Enhance CSP capabilities to support trusted types (#446)

ChrisFWood:


src/trusted-types.js
```diff
@@ -0,0 +1,26 @@
...
+function createPolicy() {
+    if (typeof trustedTypes !== "undefined") {
+        try {
+          return { policy: trustedTypes.createPolicy("esmoduleshims", policyOptions), policyEnabled: true};
+        } catch {
+
+        }
+    }
+  return { policy: policyOptions, policyEnabled: false };
+}
```
guybedford: Since we do try to save bytes here might be worth flattening this:

```suggestion
export let policy;
if (typeof trustedTypes !== "undefined") {
  try {
    policy = trustedTypes.createPolicy("esmoduleshims", policyOptions);
  } catch {}
}
```

And then replacing `policyEnabled` with just `if (policy)` kind of thing.

Also I think we need to change the implementations of `trustedInnerHTML` and `trustedScript` to work when there is no policy too, see comments below.

guybedford (commented):
Great work thank you for working on this. Can we not make the policy name `es-module-shims` or do policy names not like `-` characters? I think I would prefer that as the project name being used.

Just a couple of small comments. Then it just needs a Prettier run as well.

ChrisFWood:
> Thank you so very much for working on this.

No problem happy to contribute.

> This line fails with the error `es-module-shims-2.7.0.min.js:3160 This document requires 'TrustedScript' assignment. An HTMLScriptElement was directly modified and will not be executed.`
> 
> ```
> if (optionsScript) optionsScript.innerHTML = trustedInnerHTML(JSON.stringify(shimModeOptions));
> ```
> 
> This only happens when I set options like this:
> 
> ```
> <script type="esms-options">
> {
>   "polyfillDisable": ["css-modules", "json-modules"]
> }
> </script>
> ```
> 
> If I use the alternative:
> 
> ```
> <script>
> window.esmsInitOptions = {
>   polyfillDisable: ["css-modules", "json-modules"]
> }
> </script>
> ```
> 
> there's no error.

I might need to set up a test for this, but my guess is it's because I createHTML not createScript, my local tests all run if I swap the two over, can I ask a favour can you retest with the script version i.e.
if (optionsScript) optionsScript.innerHTML = maybeTrustedScript(JSON.stringify(shimModeOptions));

guybedford (approved):
LGTM.

Let's just maybe add a readme section or append a paragraph to the existing CSP docs noting how this works.

guybedford:
Throwing it into Claude, it suggested this for docs:

```
## Trusted Types Support

ES Module Shims supports [[Trusted Types](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API)](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API), a browser security feature that prevents DOM-based XSS attacks. When your CSP includes `require-trusted-types-for 'script'`, the library automatically creates a policy named `es-module-shims` to handle its internal script creation and source rewriting. Browsers without Trusted Types support fall back to standard behavior automatically.

To enable it, add `es-module-shims` to your CSP's trusted-types directive. If you see errors about "TrustedScript assignment," you're missing this directive.

``html
<meta http-equiv="Content-Security-Policy" 
      content="require-trusted-types-for 'script'; trusted-types es-module-shims;">
``
```

Not sure if the failure in older Chrome and Firefox is a flake or not, otherwise this might need looking into as well.

ChrisFWood:
I've no idea why it runs with trusted types but not without in that browser versions, bug in chrome I guess, I've tried manual mutation the maybeTrusted functions and the csp tests fail with the changes, but other requirements for trusted types may be hidden due to the default added to the tests.

guybedford:
Thanks, landed and released in 2.8.0.

PR #549, opened 2026-07-09
Skip feature detection in shim mode

pttsant:
In shim mode, the feature detection results are unused — the shim loader processes all modules regardless of native support. Skip the iframe-based detection entirely to avoid unnecessary DOM overhead.

### DIFF
--- a/src/features.js
+++ b/src/features.js
@@ -8,6 +8,7 @@ import {
   wasmInstancePhaseEnabled,
   wasmSourcePhaseEnabled,
   hasDocument,
+  shimMode,
   version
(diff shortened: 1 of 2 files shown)


pttsant:
@guybedford If you get a chance to review. Thanks.

guybedford (approved):

pttsant:
@guybedford Thanks for the quick merge! Could you please publish a patch to npm with this change?

pttsant:
@guybedford bumping my message from last week in case you missed it. Any concerns with publishing a patch?

PR #551, opened 2026-07-26
feat: update Wasm WPT suite to upstream, match latest source phase brand check

guybedford:
Syncs the vendored WebAssembly WPT suite in `test/` against [web-platform-tests master](https://github.com/web-platform-tests/wpt/tree/master/wasm/webapi/esm-integration) and updates the polyfill where upstream behaviour has changed.

## Polyfill change

`AbstractModuleSource.prototype[Symbol.toStringTag]` (the source phase brand check installed in `src/core.js`) now returns `undefined` for objects that are not module sources, instead of throwing a `TypeError`.

This matches upstream WPT [`wasm/jsapi/module/moduleSource.tentative.any.js`](https://github.com/web-platform-tests/wpt/blob/master/wasm/jsapi/module/moduleSource.tentative.any.js), which was changed in wpt [`c2f5b77`](https://github.com/web-platform-tests/wpt/commit/c2f5b772) — *"wasm: fix moduleSource toStringTag brand check to conform with spec undefined value"* — and the source phase imports spec, where the getter returns `undefined` rather than throwing when `HostGetModuleSourceName` is abrupt.

`test/polyfill.js` (which adapts that WPT file into the mocha suite) is updated to `assert.equal(toStringTag.call({}), undefined)`. As a side effect this also fixes the old assertion, which called the non-existent `assert.fail` and therefore passed vacuously via its own `TypeError`.

### js-string builtins

The Wasm ESM integration spec compiles modules with the `wasm:js-string` builtins enabled, so `src/core.js` now passes the `js-string` compile option through to `WebAssembly.compileStreaming`, and a new `src/wasm-js-string.js` provides a JS implementation of the builtin namespace for engines without native builtins support (ported from the [reference polyfill](https://github.com/WebAssembly/js-string-builtins/blob/main/test/js-api/js-string/polyfill.js), kept minimal for footprint — the `(array (mut i16))` accessors that can only be written in Wasm are lazily compiled from an 87 byte embedded module on first use).

Since source phase modules are instantiated by user code, the namespace is merged underneath any user-provided imports on `WebAssembly.instantiate` / `WebAssembly.Instance` when the source phase is being polyfilled. Native builtins take precedence where supported, and names not provided by the builtin namespace still resolve from the import object, matching the native fallback that the WPT test's `newbuiltin` relies on. Instance phase imports of `wasm:js-string` are no longer emitted as module resolutions and are given the namespace directly.

Footprint: 14,247B → 14,837B Brotli.

## Test suite sync

* Added `test/test-sp-source-phase-import-non-UTF8.tentative.html` + `test/resources/non-UTF8.wasm` (wpt [`526531d1`](https://github.com/web-platform-tests/wpt/commit/526531d1)) — covers source phase import of a Wasm binary that is not valid UTF-8, which exercises the polyfill's binary-safe `compileStreaming` source path.
* Added `test/test-sp-reuse-wasm-loaded-as-classic-script.tentative.html` (wpt [`526531d1`](https://github.com/web-platform-tests/wpt/commit/526531d1)) — the same Wasm URL loaded first as a classic script and then source phase imported.
* Added `test/test-sp-source-phase-js-string.tentative.html` + `test/resources/js-string-module.wasm` — source phase import of a module using the `wasm:js-string` builtins. Upstream's module is typed with non-nullable `(ref extern)` results, which need typed function references (Chrome 119+, Firefox 120+), so the import is gated on a validation probe and skipped on the older browsers in the matrix.
* Resynced `test/resources/worker-helper.js` and `test/resources/worker-source-phase.js` to upstream (wpt [`8e1b3aaa`](https://github.com/web-platform-tests/wpt/commit/8e1b3aaa)), keeping the vendored resources consistent even though the worker tests themselves are not run here.

All new tests are picked up automatically by the `test:#` chomp task. `chomp format` and `chomp build` pass.

## Deliberately left out

* **The unwrapped `WebAssembly.Global` export expectations.** `test-sp-exported-names` and `test-sp-wasm-js-cycle` still assert `typeof glob === 'number'` rather than upstream's `instanceof WebAssembly.Global`. This is the intentional deviation from #483 (`feat: support unwrapped WebAssembly.Global exports`): the esm-integration spec's `ExecuteModule` initialises global export bindings to `ToJSValue(global_read(...))`, and notes that *"it is only in the case of importing Wasm from JS that WebAssembly.Global unwrapping is observable"*. Upstream WPT is stale here, and Node's implementation also unwraps. Reverting these would regress the polyfill away from the spec.
* **The reference-types resource updates** (wpt [`140f5c80`](https://github.com/web-platform-tests/wpt/commit/140f5c80)): `exported-names.wasm` and `wasm-import-table.wasm` switch their table type from `funcref` to `externref`, and `wasm-js-cycle.wasm` gains a `mutateTab` export using `ref.func`. Pulling these in would also let the commented-out `mutateTab` assertions in `test-sp-wasm-js-cycle` be restored. They need the reference-types proposal (Chrome 96+, Firefox 79+), and the Chrome CI matrix floor for TLA-capable builds is `913920`, which is an M95 dev build — so syncing these binaries breaks that job until the matrix floor is raised. I left `.github/` alone as instructed; happy to land this as a follow-up alongside a matrix bump.
* **CSP tests** (`script-src-allows-wasm`, `script-src-allows-source-phase-wasm`, `script-src-blocks-wasm.sub`, `source-phase-blocked-by-csp`): these set `script-src 'self' 'unsafe-inline'`, which blocks the blob URLs and `unsafe-wasm-eval` the polyfill requires, so they test the CSP behaviour of the native loader rather than the polyfill. `test-csp.html` covers the polyfill's own CSP story.
* **`source-phase-preload.tentative.html`**: asserts exactly one network fetch for a dynamically inserted `<link rel=modulepreload>` of a `.wasm`. Under the polyfill the preload link is not a module the browser can resolve, and the polyfill does its own fetch, so this measures something the polyfill cannot currently satisfy.
* **Worker and worklet tests** (`worker`, `worker-import`, `worker-import-source-phase`, `worklet-import-source-phase.https`): es-module-shims cannot inject itself into a `new Worker(url, { type: 'module' })` realm, and the worklet test additionally needs HTTPS. Related: #535.

The browser test matrix was not run for this change.

Closes #534


guybedford:
> source-phase-js-string.tentative.html: requires native wasm:js-string builtins (Chrome 130+, Firefox 134+), below the browser matrix floor.

Add this, and include a polyfill for js-string builtins like the one in [the spec repo.](https://github.com/WebAssembly/js-string-builtins/blob/main/test/js-api/js-string/polyfill.js). Use maximal size optimizations to reduce footprint.

guybedford:
Failing in Firefox, we need to add the actual `builtins: ['js-string']` compilation argument for modern Firefox to use the native implementation when available. We must also set and polyfill the native strings constant to `wasm:js/string-constants`, expanding imports to it as polyfilled imports, while also passing `importedStringConstants: 'wasm:js/string-constants'` to compile.

guybedford:
Retry

guybedford:
Retry

